PR #13, opened 2023-07-13
Cli and config parsing

LDprg:
In order to start the rewrite, I did just some cli and config file parsing. Nothing special.

### DIFF
--- a/winapps-cli/Cargo.toml
+++ b/winapps-cli/Cargo.toml
@@ -2,6 +2,7 @@
 name = "winapps-cli"
 version = "0.1.0"
 edition = "2021"
+default-run = "winapps-cli"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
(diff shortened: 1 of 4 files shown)


oskardotglobal:
Looks good so far, but why are you converting the string literal to a string when creating default values for the constructor using `#[new(value = "\"RDPWindows\".to_string()")]`. Couldn't you use `#[new(value = "\"RDPWindows\"")]`

LDprg:
It is a rust thing, it comes from the difference between &str and String. The default values are &str and the struct fiels are Strings, so that stuff works out I need to convert &str to String. The fact that the conversion happen in a string and not as code is the way the default makro works (looked this up on the github page)

oskardotglobal (approved):
Aight, was just asking since I'm not that familiar with rust. If thats not problematic, I'll approve this

LDprg:
Just added some more error checking. I merge it as soon as the rust checks are ready.

PR #14, opened 2023-07-15
Update RDPApps.reg link

Leonardo-DGS:


oskardotglobal:
Thank you for your contribution. However, we can't merge this yet due to legal issues we're currently resolving.

docs/KVM.md
```diff
@@ -135,7 +135,7 @@ Leave everything as default and click `Next` through the installer. This will in
 
 ![](kvm/25.png)
 
-Once you finish the driver install, you will need to make some registry changes to enable RDP Applications to run on the system. Start by downloading the `RDPApps.reg` file from the WinApps repo by visiting https://github.com/Fmstrat/winapps/blob/main/install/RDPApps.reg, right clicking on the `Raw` button, and clicking on `Save target as`.
+Once you finish the driver install, you will need to make some registry changes to enable RDP Applications to run on the system. Start by downloading the `RDPApps.reg` file from the WinApps repo by visiting https://github.com/winapps-org/winapps/blob/main/install/RDPApps.reg, right clicking on the `Raw` button, and clicking on `Save target as`.
```
fbartels, line 138: Can we make this link relative?

fbartels (changes requested):

oskardotglobal (approved):
Looks good

LDprg:
Sure a relative link would be copied correctly when right clicking -> Copy? Because many people just copy the link that way and past it into the vm.

fbartels (approved):

fbartels:
@LDprg no, this would not work for a direct paste into curl. But the behaviour with a relative or absolute link still give the same result (a Github page). The readme also explains what do it. So while not optimal, i think there is enought guidance on what to do.

Leonardo-DGS:
> Sure a relative link would be copied correctly when right clicking -> Copy? Because many people just copy the link that way and past it into the vm.

I've tried and it works.

oskardotglobal:
Can confirm. Merging this since relicensing isn't an issue, see: https://github.com/winapps-org/winapps/discussions/11#discussioncomment-6461682

oskardotglobal:
@Leonardo-DGS Thank you for your contribution

PR #27, opened 2023-07-27
[ci skip] Skip rust-check and clippy on pre-commit.ci

oskardotglobal:


LDprg (approved):
Well not much to say. Ready to merge.

PR #30, opened 2023-07-30
Adding freerdp backend

LDprg:
Still WIP. 

Feel free to discuss the code (for now it isn't that much).


oskardotglobal:
@CLAassistant recheck

winapps/src/lib.rs
```diff
@@ -33,16 +43,6 @@ pub struct RemoteConfig {
...
-    }
-
-    fn load_config(&self, path: &str);
-
-    fn run_app(&self, app: &str);
-}
-
 pub fn load_config(path: Option<&str>) -> Config {
     let home = home_dir().expect("Could not find the home path!");
     let default = &format!("{}{}", home.to_str().unwrap(), "/.config/winapps/");
```
AkechiShiro, line 48: I believe we should try and get an environment variable and default that PATH to it if it is set, the environment variable is : `XDG_CONFIG_DIRS`, this would cover an edge case, where a user has all his configuration elsewhere than in `$HOME/.config`.

> $XDG_CONFIG_HOME defines the base directory relative to which user-specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

But I'm not sure a fix should be in your PR on this matter, I'll open up another PR for this.

LDprg: Oh thx, definitely something I will change as soon as I have time.

AkechiShiro: If you believe you won't be able before a few months, feel free to let me know, I'll close this PR and build from your branch a new one.
I'd like to move this project forward, thanks for your initial draft !

LDprg:
Merged it to move the development to its own branch so we can make working on this easier.

PR #42, opened 2023-09-10
Add freerdp backend

LDprg:
This moved from #30 to here for better way of working.

### DIFF
--- a/winapps/src/freerdp.rs
+++ b/winapps/src/freerdp.rs
@@ -14,7 +14,7 @@ pub mod freerdp_back {
             xfreerdp.spawn().expect("Freerdp execution failed! It needs to be installed!");
             println!("Freerdp found!");
             
-            println!("Checks success!");
+            println!("All dependencies found!");
         }
 
         fn run_app(&self, config: Config, _app: &str) {
--- a/winapps/src/freerdp.rs
+++ b/winapps/src/freerdp.rs
@@ -33,7 +33,6 @@ pub mod freerdp_back {
                 Some(exe) => {
                     xfreerdp.args([
                         &format!("/app:{}", exe),
-                        "-wallpaper",
                         &format!("/d:{}", &config.rdp.domain),
                         &format!("/u:{}", &config.rdp.username),
                         &format!("/p:{}", &config.rdp.password),
--- a/winapps/src/freerdp.rs
+++ b/winapps/src/freerdp.rs
@@ -29,28 +29,28 @@ pub mod freerdp_back {
             match app {
                 Some(exe) => {
                     xfreerdp.args([
-                        "/app:".to_owned() + exe,
-                        "-wallpaper".to_owned(),
-                        "/d:".to_owned() + &config.rdp.domain,
-                        "/u:".to_owned() + &config.rdp.username,
-                        "/p:".to_owned() + &config.rdp.password,
-                        "/v:".to_owned() + &config.rdp.host,
-                        "/dynamic-resolution".to_owned(),
-                        "+auto-reconnect".to_owned(),
-                        "+clipboard".to_owned(),
-                        "+home-drive".to_owned(),
+                        &format!("/app:{}", exe),
+                        "-wallpaper",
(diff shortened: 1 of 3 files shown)


winapps/src/freerdp.rs
```diff
@@ -19,19 +19,24 @@ pub mod freerdp_back {
...
             println!("Test finished!");
         }
 
-        fn run_app(&self, config: Config, _app: &str) {
+        fn run_app(&self, config: Config, app: &str) {
             let mut xfreerdp = Command::new("xfreerdp");
             xfreerdp.args([
-                //"/app:".to_owned() + app,
+                "/app:".to_owned() + app,
                 "/d:".to_owned() + &config.rdp.domain,
```
oskardotglobal, line 31: You should probably use [`format!`](https://doc.rust-lang.org/std/macro.format.html) for these.
Also, you can pass `&str` to these instead of `String`, see https://github.com/oskardotglobal/winapps/blob/rewrite/winapps/src/quickemu.rs#L34-L42

oskardotglobal (changes requested):

winapps/src/freerdp.rs
LDprg: Thx, forgot about this one.

LDprg:
So I think generally the the base is finished and ready for merge. Just waiting for some reviews.

Maybe we want to change to command names (later this would we much more complicated).

oskardotglobal (approved):
Looks great.

oskardotglobal:
> Maybe we want to change to command names
We could just keep what you've called `app` as `manual`, but imo `run` would be better

AkechiShiro:
@oskardotglobal did you try and build this branch ? Or should I test it on my side ?

oskardotglobal:
Nope, I'll test this later

AkechiShiro:
I tested the build works, but there are a few issues, I noticed @LDprg

- `winapps-cli check` image
- I tried a few commands, but they seemed to fail if `XDG_CONFIG_HOME` wasn't defined, we should automatically fallback on `$HOME/.config`

oskardotglobal:
@AkechiShiro the latter is included in #26

LDprg:
@AkechiShiro The first one is that winapps doesnt disable the freerdp output (which containst alot of unnecessary warings). I will investigate into this.

LDprg:
@AkechiShiro did you test it with the generated quickemu from #26? I am not sure if we need a port forwarding from the guest to host to make rdp communication working per default.

AkechiShiro:
@LDprg no I removed all the previous files to test cleanly but I understand that there is no quickemu get and VMs files

oskardotglobal:
@LDprg I'll test it

LDprg:
So finally I fixed the log outputs of freerdp causing unessecary warnings to be shown. Later we could redirect its output to some log files, however this will be in a sperate PR (just want this get through as fast as possible). 

Furthermore I fixed some minor issues after the merge of #26 and renamed the app command to run.
@oskardotglobal can you review this so I can merge.

winapps/src/freerdp.rs
```diff
@@ -6,20 +6,58 @@ pub mod freerdp_back {
...
-            todo!()
+        fn run_app(&self, config: Config, app: Option<&String>) {
+            let mut xfreerdp = Command::new("xfreerdp");
+            xfreerdp.stdout(Stdio::null());
+            xfreerdp.stderr(Stdio::null());
+            match app {
+                Some(exe) => {
+                    xfreerdp.args([
+                        &format!("/app:{}", exe),
+                        "-wallpaper",
```
oskardotglobal, line 36: Any reason why you add the `-wallpaper` flag here and not if the app isn't given?

oskardotglobal (changes requested):
Looks good otherwise

LDprg:
@oskardotglobal I removed it. Can you approve so I can merge?

oskardotglobal (approved):
Looks good.

PR #48, opened 2023-10-07
feat: error handling

oskardotglobal:
All those `.expect`s and nested match statements were bugging me, so I've written a custom error handler.

Basically, I've written 2 macros, which do the following:
- `unwrap_or_panic!` & `unwrap_or_exit!` - Unwrap either an Option or a Result. Log the error of the result if it exists + further context with `format!` syntax support
- `error!` & `error_from!` - Construct a `WinappsError` from an existing error and / or a message. This is mostly used to return values from functions, but you can still call `panic` and `exit` on it for the functionality of the 2 macros above 

I've also documented all of them and all the structs, enums and traits I used.
This also includes the commits from [#26](https://github.com/winapps-org/winapps/pull/26), so that has to be merged first.


oskardotglobal:
This is ready for review now with #26 being merged

LDprg:
@oskardotglobal seems like something messed up with your repo, because it for example readds the quickemu installer script.

oskardotglobal:
Yeah, gotta merge first

oskardotglobal:
Ok, this is fully up-to-date now.

winapps/src/quickemu.rs
```diff
@@ -1,89 +1,68 @@
...
-        data_dir.join(format!("{}/{}.pid", config.vm.short_name, config.vm.name)),
-    ) {
-        Ok(pid) => {
-            let pid = pid.trim();
+    let pid = unwrap_or_exit!(
+        fs::read_to_string(
+            data_dir.join(format!("{}/{}.pid", config.vm.short_name, config.vm.name)),
+        ),
+        "Failed to read PID file, is the VM running and the config correct?"
+    );
```
LDprg, line 60: Is there a reason you removed the trim function?

winapps/src/quickemu.rs
```diff
@@ -1,89 +1,68 @@
...
-        Err(e) => {
-            println!("Failed to execute quickemu: {}", e);
-            println!("Please make sure quickemu is installed and in your PATH");
-            exit(1);
-        }
-    };
+    let command = unwrap_or_exit!(
+        Command::new("quickemu")
+            .current_dir(data_dir)
+            .args([
```
LDprg, line 33: Seems like you forgot the ignore msrs

LDprg (changes requested):
Some minor code parts did get lost, otherwise great work.

winapps/src/quickemu.rs
oskardotglobal: Nope, I missed that

oskardotglobal:
Ok I've made some more changes; I've fully removed the `unwrap_or_exit` and `unwrap_or_panic` functions because them having the same name as the macros was kind of stupid because the IDE would automatically import those instead of the macros
I've replaced them with the `into_error` macro which just wraps the result in a function that does some type magic and returns a `Result<T, WinappsError>`. I've also documented all my changed

winapps/src/quickemu.rs
LDprg: Still not the same, you only trim config.vm.name.

LDprg (changes requested):
Almost finished.

oskardotglobal:
there we go

LDprg (approved):
This is now ready.

PR #56, opened 2023-10-28
feat: basic ci

oskardotglobal:
Basic CI that builds winapps, just 2 binaries for now; an x86 linux binary and an experimental aarch64 one (built on musl, should have excellent support anywhere)
We could probably remove the aarch64 one, but we could also keep it for now
The way this works is once a release is created, the action will automatically attach the files + sha256 hashes to the release


AkechiShiro:
Is there anything, that I should do to help with this merge? I believe the CI action is alright for now.

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+          - os_name: Linux-aarch64
+            os: ubuntu-20.04
+            target: aarch64-unknown-linux-musl
+            bin: winapps-linux-arm64
+          - os_name: Linux-x86_64
+            os: ubuntu-20.04
+            target: x86_64-unknown-linux-gnu
+            bin: winapps-linux-amd64
+        toolchain:
+          - stable
```
LDprg, line 28: Sure that everything builds on stable rust? It is quite common to use nightly (I do use the nightly toolchain) and I did not check if everything works on stable.

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+      matrix:
+        platform:
+          - os_name: Linux-aarch64
+            os: ubuntu-20.04
+            target: aarch64-unknown-linux-musl
+            bin: winapps-linux-arm64
+          - os_name: Linux-x86_64
+            os: ubuntu-20.04
+            target: x86_64-unknown-linux-gnu
+            bin: winapps-linux-amd64
```
LDprg, line 26: Theoretically we could provide x86 binaries for older system that uses winapps for remote working. (VMs propably don't run on these kind of systems)

oskardotglobal: What do you mean by "older systems"

.github/workflows/build.yml
oskardotglobal: I use stable; it wouldn't be a bad idea to use nightly though, what do ya'll think? @AkechiShiro

.github/workflows/build.yml
LDprg: 32bit Computers

.github/workflows/build.yml
AkechiShiro: I prefer nightly as well, is there any risk for nightly to break build a lot of times ? Or maybe a better question is do we need nightly features ?

.github/workflows/build.yml
AkechiShiro: That's pretty old, I don't think they're sold anymore but that would be nice if it's possible and not too much of a hassle to maintain.

.github/workflows/build.yml
oskardotglobal: I don't think so, and we can rebuild any time so it shouldn't be a big deal

.github/workflows/build.yml
oskardotglobal: Yeah, I think that would be a good idea

oskardotglobal:
Alright, dropping aarch64 in favour of i686 (this seems to be the 32-bit variant afaik) because freerdp seems to not actually support aarch64 from what I've seen

oskardotglobal:
I also added musl binaries just in case; especially thin clients will profit from this

oskardotglobal:
oh also because we don't have any packages yet I added a curl-bash script for installation; it will also install quickemu using the other script

LDprg (approved):
Seems good. In future we may replace the install.sh with a python script, in order to uniform things a bit (So we have rust for compiled and python for scripting).

PR #61, opened 2023-11-10
Make it clear IP address is required if not using KVM

### LINKED ISSUE #60: Clarify that an IP address is required if not using QEMU
I'd much rather run Windows on my dedicated Proxmox server than using up resources on my primary gaming PC, which as I understand it, should be possible as winapps is basically a wrapper on freerdp to make it easier to set up. But with my config file set up, running `winapps check` or `installer.sh` stall immediately because I'm not a member of `libvirt` or `kvm` groups. I have no desire to virtualize using my desktop's resources, so I don't think I would need libvirt/kvm installed, but I don't know how to get the rest of the setup finished without it.

Hyperspeed1313:
Closes #60

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ RDP_PASS="MyWindowsPassword"
 The username and password should be a full user account and password, such as the one created when setting up Windows or a domain user. It cannot be a user/PIN combination as those are not valid for RDP access.
 
 Options:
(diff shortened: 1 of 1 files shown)


oskardotglobal (approved):
Looks good. Merging

PR #88, opened 2024-04-20
Prefer to use Theme Icons

GreatNovaDragon:
Some iconpacks like Papyrus do come with their own icons for applications that are seemingly also just windows only. It is almost always preferrable to use the iconpack instead of the own svgs. 

I added an ICON field to the app-info where one can point to the icon-name.


oskardotglobal (approved):
Looks good.

PR #94, opened 2024-05-14
Add docker documentation for legacy winapps 

LDprg:
Docker seems a better option for hosting the vm.

Thid Pr adds documentation on how to use docker with legacy winapps.


oskardotglobal:
This is my compose file, in case you need it
```yaml
name: "winapps"

volumes:
  data:

services:
  windows:
    image: dockurr/windows
    container_name: windows
    environment:
      VERSION: "core11"
      RAM_SIZE: "8G"
      CPU_CORES: "4"
    devices:
      - /dev/kvm
    cap_add:
      - NET_ADMIN
    ports:
      - 8006:8006
      - 3389:3389/tcp
      - 3389:3389/udp
    stop_grace_period: 2m
    restart: on-failure
    volumes:
      - data:/storage
```

oskardotglobal:
The only step I had to take was to manually install the .reg file, but that can be automated

LDprg:
The only change I would make is to set privileged to true and remove the device/capp_add section. There could be some trouble about permissions if not done so (on my system for example docker wasn't able to access my network interface). It would also simplify gpu passthrough (since devices do not need to be listed twice) at the cost of some not that important security.

oskardotglobal:
Make sure to mention that the RDP_IP has to be set to `localhost` or `127.0.0.1` (which might lead to better compatibility) in the config explicitely

oskardotglobal:
Also, we might wanna use bind mounts instead of named volumes by default

LDprg:
> Also, we might wanna use bind mounts instead of named volumes by default

I feel like we should do it the other way around. Use volumes as default and mention bind mounts (or maybe even recommend). Since Volumes are more independent and easier to setup on a just works system.

oskardotglobal:
My reason behind bind mounts would be to allow access to the guest's fs easily

oskardotglobal:
or wait, the mount contains a disk image, doesn't it
yeah then volumes are better

LDprg:
In general this is ready for review. I will also do a quick rebase later.

docs/docker.md
```diff
@@ -0,0 +1,67 @@
...
+
+# Requirements
+
+Since docker manages the dependencies of the container automatically you only need to install docker or podman itself. (Podman is recommended because of the faster container startup times. Note that podman and docker are interchangeable so no instructions will change depending on which one you use)
+
+Note: This will only work on linux systems since some kernel interfaces (like kvm) are needed by the vm. Because of this performance can vary in kernel versions (newer will likely perform better).
+
+# Setup docker container
+
+The easiest way to setup a windows vm is by using docker compose. Just create a `docker-compose.yml` with following content:
```
oskardotglobal, line 15: The preferred name for the compose file is `compose.yaml` now, see https://docs.docker.com/get-started/08_using_compose/

oskardotglobal: I see you're recommending podman as the runtime here, then we should definitely call the file `compose.yaml`

docs/docker.md
```diff
@@ -0,0 +1,67 @@
+# Docker
+
+## Why docker?
+
+While working with virsh is completely fine for winapps, however you have to setup and optimise you vm manually. Docker on the other hand setups most of the stuff automatically and also makes the vm highly portable between linux distros.
+
+# Requirements
+
+Since docker manages the dependencies of the container automatically you only need to install docker or podman itself. (Podman is recommended because of the faster container startup times. Note that podman and docker are interchangeable so no instructions will change depending on which one you use)
```
oskardotglobal, line 9: For compose to work with podman the podman socket has to be enabled which isn't the case by default on most distros

docs/docker.md
```diff
@@ -0,0 +1,67 @@
...
+      - 3389:3389/udp
+    stop_grace_period: 2m
+    restart: on-failure
+    volumes:
+      - data:/storage
+```
+
+Now you can tune the ram/usage by changing RAM_SIZE/CPU_CORES. You can also specify the windows versions you want to use. You might also want to take a look at the [docker image repo](https://github.com/dockur/windows).
+
+Note: Older versions than Windows 10 are not officially supported. However they might still work with some additional tuning.
```
oskardotglobal, line 44: Use a callout here

docs/docker.md
```diff
@@ -0,0 +1,67 @@
...
+services:
+  windows:
+    image: dockurr/windows
+    container_name: windows
+    environment:
+      VERSION: "win11"
+      RAM_SIZE: "8G"
+      CPU_CORES: "4"
+    privileged:
true
```
oskardotglobal, line 31: Is there a special character here that shouldn't be? The github web editor shows a `^M` here

docs/docker.md
```diff
@@ -0,0 +1,67 @@
...
+
+volumes:
+  data:
+
+services:
+  windows:
+    image: dockurr/windows
+    container_name: windows
+    environment:
+      VERSION: "win11"
```
oskardotglobal, line 28: I think we should ship tiny11 by default

docs/docker.md
oskardotglobal: Also, maybe link the official docker/podman installation guides

oskardotglobal (changes requested):
Looks mostly good

docs/docker.md
LDprg: I wasnt able to find this character. On my system the we editor just shows everything fine.

LDprg: Ah ok opened on another device and got that char visible. Fixed

LDprg:
All stuff should be fixed

oskardotglobal (approved):
Looks good.

Issue #91, opened 2024-04-29
Latest FreeRDP Dependency Behavior and Documentation


Ahmedsaed:
**Describe the bug**
The behavior of the FreeRDP dependency differs from user expectations due to the discrepancy between the version packaged in certain Linux distributions and the version marked as latest by the developer. Updating the documentation or providing a warning within the application may help guide users encountering this issue.

**Steps to Reproduce**:
1. Install FreeRDP from the repository of a Linux distribution that packages the latest released version (e.g., version 3.4.0).
1. Attempt to use FreeRDP with the `/app` argument.
1. Observe FreeRDP printing the help message and exiting unexpectedly.

**Expected Behavior**:
FreeRDP should properly handle the `/app` argument and perform the intended operation without issues.

**Actual Behavior**:
In the latest couple of released versions of FreeRDP (e.g., version 3.4.0), using the `/app` argument causes FreeRDP to print the help message and exit instead of executing the desired operation.

**System information:**
- Linux Distribution: Fedora 40
- Version of FreeRDP: 3.4.0

**Impact**:
This issue affects users relying on FreeRDP distro packages, particularly those needing to use the /app argument.

**Workaround**:
- Compiling the correct version of FreeRDP
- Using Distrobox or similar solutions to run a Linux container that has the binaries for the correct version in the repositories.
    - Use `distrobox-export` to export the binary to host system. Ex: `distrobox-export --bin /usr/bin/xfreerdp ~/.local/bin`

KernelGhost:
@Ahmedsaed Could you please confirm if this issue still persists? WinApps should now fully support FreeRDPv3 and also supports FreeRDP installed as a Flatpak for users on distributions that don't offer version 3 via the official package manager repositories. I am running FreeRDP 3.6.2 and have not encountered any issues with the `/app` flag.

f8ith:
I can reproduce the issue. The issue happens because winapps loads the config (Adds the flags to $RDP_COMMAND) before scanning for the system's FreeRDP version, causing the $FREERDP_COMMAND to be fixed as xfreerdp which is the default.
On systems where the binary is named "xfreerdp3", the error occurs. 

Simply moving the waGetFreeRDPCommand before waLoadConfig fixes the issue.

PR #98, opened 2024-05-23
Precommit for legacy

LDprg:
This is still wip and will need extensive testing when ready.

Since a there is a lot of formatting and manual fixing.


oskardotglobal:
pre-commit.ci autofix

LDprg:
This is basically ready for early review and testing. Note that I DIDN'T TEST THIS, so it shouldn't be right up merged.

LDprg:
Btw for testing I can really recommend using distrobox. (I know docker in docker is kinda strange but it works). It allows us to create a clean environment and test different distros without installing a vm for like 20 minutes.

oskardotglobal:
Ok I'm having quite a bit of trouble getting winapps to work; it just refuses on my Nix machine
The problem is that both the domain and the password have to be empty in my case and due to shell escaping the command winapps runs becomes
```bash
xfreerdp /p: /d: /u:Docker /scale:100 +auto-reconnect +home-drive -wallpaper +dynamic-resolution +span /app:program:explorer.exe /v:127.0.0.1
```
instead of
```bash
xfreerdp /p:"" /d:"" /u:Docker /scale:100 +auto-reconnect +home-drive -wallpaper +dynamic-resolution +span /app:program:explorer.exe /v:127.0.0.1
```
which works completely fine

oskardotglobal:
Any suggestions?

oskardotglobal (approved):
Looks good. Works on my machine.

oskardotglobal:
I've also found a fix for my above problem & included the automatic installation of the reg file in this PR too

LDprg:
@oskardotglobal can you update the docs accordingly?

oskardotglobal:
Yeah, sure

oskardotglobal:
.. Why does pre-commit remove the double spaces after lines
They represent line breaks in Markdown

oskardotglobal:
pre-commit.ci autofix

LDprg:
> .. Why does pre-commit remove the double spaces after lines
> They represent line breaks in Markdown

Propably some tab/line ending fixing precommit hook.

CiRiX-:
What about this issue here? https://github.com/winapps-org/winapps/issues/91

The newer freerdp releases wont start apps.

oskardotglobal:
@CiRiX- 
That seems like an issue with the user's specific version or with the changed syntax of the `/app` argument
I don't wanna be *that* guy, but it seems like a skill issue; I'm running freerdp 3.5.1 on my Nix machine and it works just fine

LDprg:
@oskardotglobal we might add a switch for the /app syntax depending on the freerdp version?

This would work around some package manager not providing any freerdp 3 packages.

Or we ship a working freerdp v3 command with winapps.

oskardotglobal:
I think shipping a binary would be a good approach, but that would require the binary to not be linked against OS libraries since otherwise you would have to install a bunch of dependencies

CiRiX-:
@oskardotglobal 
Don't worry, you are probably right :wink: 
I am not super experienced with Linux, and i'm not a programmer.

I think it should be more user friendly, so shipping a working version like @LDprg said would be a good idea i guess.

AkechiShiro:
Maybe we'd need to work a wrapper for Xfreerdp (in Bash maybe, that simplifies usage) ?

PR #107, opened 2024-05-30
Fix some bugs in installer.sh

oskardotglobal:
- `$SUDO` shouldn't be quoted as that causes command not found errors
- the loop on ln. 50 didn't actually use the loop variable but specified a bad full path

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -42,7 +42,7 @@ function waFindInstalled() {
         rm -f "$HOME/.local/share/winapps/installed"
         rm -f "$HOME/.local/share/winapps/detected"
         cp "$DIR/install/ExtractPrograms.ps1" "$HOME/.local/share/winapps/ExtractPrograms.ps1"
-        for F in $DIR/apps/*; do
+        for F in "$DIR"/apps/*; do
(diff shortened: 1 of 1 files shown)


installer.sh
```diff
@@ -124,9 +122,9 @@ function waConfigureApps() {
         APPS+=("$FULL_NAME ($F)")
         INSTALLED_EXES+=("$(echo "${WIN_EXECUTABLE##*\\}" | tr '[:upper:]' '[:lower:]')")
     done < <(sed 's/\r/\n/g' < "$HOME/.local/share/winapps/installed")
-    # FIXME
```
LDprg, line 127: This fixme is about the 2207 error. If I remember correctly itbwas about the way the sort is done.

I would either fix it or not removing the fixme.

LDprg (changes requested):
Great Pr. Just a minor not directly related issue.

LDprg (approved):
Seems good

LDprg:
precommit.ci autofix

LDprg:
pre-commit.ci autofix

LDprg:
pre-commit.ci autofix

PR #108, opened 2024-05-30
Fixing installer.sh

Alchemi1963:
Rewrote the printf lines to work nice with writing the batch script.
Rewrote the apps scanner to take relative paths (with ls) for more consistent results across
Added checks for emtpy vars so *bin* aren't accidentally written to or deleted.

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -122,7 +122,7 @@ function waConfigureApps() {
     while IFS= read -r F; do
         [[ -n $F ]] || continue
         F=$(echo "$F" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
-        echo \""$F"\" >> test
+        echo \""$F"\" >> "test"
         # shellcheck disable=SC1090
         . "$DIR/apps/$F/info"
         APPS+=("$FULL_NAME ($F)")
--- a/installer.sh
+++ b/installer.sh
@@ -44,7 +44,7 @@ function waFindInstalled() {
         cp "$DIR/install/ExtractPrograms.ps1" "$HOME/.local/share/winapps/ExtractPrograms.ps1"
         for F in "$DIR"/apps/*; do
             [[ -e "$F" ]] || break
-            F="$(basename "$F")" 
+            F="$(basename "$F")"
 
             # shellcheck disable=SC1090,SC1091
             . "$DIR/apps/$F/info"
@@ -173,7 +173,7 @@ function waConfigureAppsAllOfficiallySupported(){
     COUNT=0
     while IFS= read -r F; do
         F=$(echo "$F" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
-        
+
         COUNT=$((COUNT + 1))
         $SUDO cp -r "apps/$F" "$SYS_PATH/apps"
         waConfigureApp "$F" svg
--- a/installer.sh
+++ b/installer.sh
@@ -42,15 +42,13 @@ function waFindInstalled() {
         rm -f "$HOME/.local/share/winapps/installed"
         rm -f "$HOME/.local/share/winapps/detected"
         cp "$DIR/install/ExtractPrograms.ps1" "$HOME/.local/share/winapps/ExtractPrograms.ps1"
-        for F in "$DIR"/apps/*; do
-            [[ -e "$F" ]] || break
+        for F in $(ls "${DIR}/apps"); do
             # shellcheck disable=SC1090,SC1091
-            . "$F/info"
-            printf "IF EXIST \"%s\" ECHO %s >> \\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp" "$WIN_EXECUTABLE" "$F" >>"$HOME/.local/share/winapps/installed.bat"
+            . "$DIR/apps/$F/info"
+	    printf "IF EXIST \"%s\" ECHO %s >> \\\\\\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp\n" "$WIN_EXECUTABLE" "$F" >> "$HOME/.local/share/winapps/installed.bat"
         done
-        printf "powershell.exe -ExecutionPolicy Bypass -File \\\\tsclient\\home\\.local\\share\\winapps\\ExtractPrograms.ps1 > \\\\tsclient\home\\.local\\share\\winapps\\detected" >>"$HOME/.local/share/winapps/installed.bat"
-        printf "RENAME \\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp installed" >>"$HOME/.local/share/winapps/installed.bat"
-        # shellcheck disable=SC2140
+        printf "powershell.exe -ExecutionPolicy Bypass -File \\\\\\\\tsclient\\home\\.local\\share\\winapps\\\\ExtractPrograms.ps1 > \\\\\\\\tsclient\home\\.local\\share\\winapps\\detected\n" >> "$HOME/.local/share/winapps/installed.bat"
+        printf "RENAME \\\\\\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp installed\n" >> "$HOME/.local/share/winapps/installed.bat"        # shellcheck disable=SC2140
         $FREERDP_COMMAND /d:"$RDP_DOMAIN" /u:"$RDP_USER" /p:"$RDP_PASS" +auto-reconnect +home-drive -wallpaper +span /app:program:"C:\Windows\System32\cmd.exe",cmd:"/C \\\\tsclient\\home\\.local\\share\\winapps\\installed.bat" /v:"$RDP_IP" 1>/dev/null 2>&1 &
         COUNT=0
         while [ ! -f "$HOME/.local/share/winapps/installed" ]; do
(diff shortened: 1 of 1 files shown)


installer.sh
```diff
@@ -42,15 +42,13 @@ function waFindInstalled() {
         rm -f "$HOME/.local/share/winapps/installed"
         rm -f "$HOME/.local/share/winapps/detected"
         cp "$DIR/install/ExtractPrograms.ps1" "$HOME/.local/share/winapps/ExtractPrograms.ps1"
-        for F in "$DIR"/apps/*; do
-            [[ -e "$F" ]] || break
+        for F in $(ls "${DIR}/apps"); do
             # shellcheck disable=SC1090,SC1091
-            . "$F/info"
```
oskardotglobal, line 48: This is correct as is. Do not change this.

installer.sh
```diff
@@ -42,15 +42,13 @@ function waFindInstalled() {
...
-            printf "IF EXIST \"%s\" ECHO %s >> \\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp" "$WIN_EXECUTABLE" "$F" >>"$HOME/.local/share/winapps/installed.bat"
+            . "$DIR/apps/$F/info"
+            printf "IF EXIST \"%s\" ECHO %s >> \\\\\\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp\n" "$WIN_EXECUTABLE" "$F" >> "$HOME/.local/share/winapps/installed.bat"
```
oskardotglobal, line 48: Those four extra backslashes are invalid from what I understand, or at least shellcheck does not like them; remove them

installer.sh
```diff
@@ -42,15 +42,13 @@ function waFindInstalled() {
...
         done
-        printf "powershell.exe -ExecutionPolicy Bypass -File \\\\tsclient\\home\\.local\\share\\winapps\\ExtractPrograms.ps1 > \\\\tsclient\home\\.local\\share\\winapps\\detected" >>"$HOME/.local/share/winapps/installed.bat"
-        printf "RENAME \\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp installed" >>"$HOME/.local/share/winapps/installed.bat"
-        # shellcheck disable=SC2140
+        printf "powershell.exe -ExecutionPolicy Bypass -File \\\\\\\\tsclient\\home\\.local\\share\\winapps\\\\ExtractPrograms.ps1 > \\\\\\\\tsclient\home\\.local\\share\\winapps\\detected\n" >> "$HOME/.local/share/winapps/installed.bat"
+        printf "RENAME \\\\\\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp installed\n" >> "$HOME/.local/share/winapps/installed.bat"        # shellcheck disable=SC2140
```
oskardotglobal, line 51: That shellcheck directive needs to be on the next line

oskardotglobal: Also, same thing with the backslashes

oskardotglobal (changes requested):
Thank you for this pull request. Just some minor issues

installer.sh
jrevillard: There are needed... I fixed everything like this yesterday night too 
`\n` i also mandatory.

installer.sh
jrevillard: There are needed... I fixed everything like this yesterday night too

oskardotglobal:
I will just test this on my machine and report back

jrevillard:
> I will just test this on my machine and report back

I have another fix also... I did not made any MR but here is my diff: 
[installer.PATCH](https://github.com/user-attachments/files/15509914/installer.PATCH)

It's still not perfect... I have the same problem than #111 and  #110 (this later I simply removed the check as you see)

installer.sh
LDprg: Why are they needed? \\\\tsclient results in \\tsclient which seem correct to me. So why should it the double amount of backslashes?

jrevillard: because `\t` is also interpreted as a tabulation :smile:

Alchemi1963: Yes, @jrevillard got it right.
The four backslashes got interpreted as one.
So I added four more, if the shellcheck doesn't like them the script should probably go back to using echo instead of printf or the shellcheck needs to be disabled for that line (I think that's possible? Not sure though, never used it before).

oskardotglobal:
I have a fix for the printf problem, gonna commit that rq

Alchemi1963:
> > I will just test this on my machine and report back
> 
> I have another fix also... I did not made any MR but here is my diff: [installer.PATCH](https://github.com/user-attachments/files/15509914/installer.PATCH)
> 
> It's still not perfect... I have the same problem than #111 and #110 (this one I simply removed the check as you see)

I see you did mostly the same things and had the same issue as I in #109 :)
I did that too, but didn't commit it as the devs specifically made a commit a while back about preferring system icons instead (idk why but ok)

Alchemi1963:
> I have a fix for the printf problem, gonna commit that rq

Out of curiosity, what is that fix?

oskardotglobal:
@Alchemi1963 

If you run a printf like
```shell
printf "%s\n" 'powershell.exe -ExecutionPolicy Bypass -File \\tsclient\home\.local\share\\winapps\ExtractPrograms.ps1 > \\tsclient\home\.local\share\winapps\detected'
```
with the argument containing backslashes in single quotes and have a first argument in double quotes then the backslashes will be interpreted literally

oskardotglobal:
No more double or triple escaping

oskardotglobal (changes requested):
Ok, this is now ready to go once the issues mentioned by shellcheck are resolved.

Alchemi1963:
> @Alchemi1963
> 
> If you run a printf like
> 
> ```shell
> printf "%s\n" 'powershell.exe -ExecutionPolicy Bypass -File \\tsclient\home\.local\share\\winapps\ExtractPrograms.ps1 > \\tsclient\home\.local\share\winapps\detected'
> ```
> 
> with the argument containing backslashes in single quotes and have a first argument in double quotes then the backslashes will be interpreted literally

Oh yeah, that is *much* cleaner

Firelfy:
This PR allows allows installer.sh to complete successfully for me. No more RDP connection error message.

LDprg (approved):
Thats very nice work. Thx for ever contributing.

LDprg:
pre-commit.ci autofix

LDprg:
We only need to fix shellcheck, before merging.

installer.sh
```diff
@@ -120,9 +121,9 @@ function waConfigureApps() {
     APPS=()
     while IFS= read -r F; do
         [[ -n $F ]] || continue
+        F=$(echo "$F" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
+        echo \""$F"\" >> test
```
oskardotglobal: Why is this still quoted twice?

installer.sh
```diff
@@ -122,7 +122,7 @@ function waConfigureApps() {
     while IFS= read -r F; do
         [[ -n $F ]] || continue
         F=$(echo "$F" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//')
-        echo \""$F"\" >> test
+        echo \""$F"\" >> "test"
```
oskardotglobal, line 125: Doesn't that just do nothing now? I think this is supposed to be a check whether F is empty

oskardotglobal (changes requested):

installer.sh
LDprg: One quote is for ensuring $F is correctly accessed. \" is for the quote in the string.

installer.sh
LDprg: @oskardotglobal It writes the content of $F into a test file. Wasn't sure if we need it but we need to double quote so shellcheck doesn't complain about misconceptions with the test command.

Not sure if we could remove it.

oskardotglobal: I thought it piped into the `test` command (https://linux.die.net/man/1/test), but that makes sense too

oskardotglobal: If I search the file there are no other occurences of `test`
I say we remove it.

LDprg: I agree

installer.sh
```diff
@@ -45,13 +45,13 @@ function waFindInstalled() {
         for F in "$DIR"/apps/*; do
             [[ -e "$F" ]] || break
             # shellcheck disable=SC1090,SC1091
-            . "$F/info"
-            printf "IF EXIST \"%s\" ECHO %s >> \\\\tsclient\\home\\.local\\share\\winapps\\installed.tmp" "$WIN_EXECUTABLE" "$F" >>"$HOME/.local/share/winapps/installed.bat"
+            . "$DIR/apps/$F/info"
```
IsaacZD, line 48: I'm experimenting this patch and I'm getting errors from this line. Seems $F is already absolute path in so "$DIR/apps" isn't really needed?

oskardotglobal: Oh, right
Thanks for mentioning it, I would've missed that

oskardotglobal: (We previously fixed that in another PR)

oskardotglobal (approved):
Alright, this looks ready to merge for me

LDprg (approved):
Seems good to me

installer.sh
```diff
@@ -46,12 +46,12 @@ function waFindInstalled() {
             [[ -e "$F" ]] || break
             # shellcheck disable=SC1090,SC1091
             . "$F/info"
```
IsaacZD, line 48: @oskardotglobal Thanks for your rapid response! I didn't expect it to be updated that fast : ).
So I've played with it a bit more and seems to have fixed errors in user mode (not sure about system as I'm not planning doing that). What I did is simply:
1. Add F="$(basename "$F")" above this line so we can use "$DIR/apps/$F/info" now. This is necessary because other places are also using $F and simply remove "$DIR/apps" in this line doesn't fix other places.
2. Add F=$(echo "$F" | sed -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//') to all "while IFS= read -r F" blocks because I'm getting errors due to trailing spaces.

Not sure if these fixes apply to others though because I'm using Nixos which is quite different to other distros, so probably needs to test in other environments.

Gonna go sleep now but hope this is helpful ; )

oskardotglobal: This is very helpful, thank you
I'm actually using Nix too, but these seem like general issues so I've commited the fix

PR #116, opened 2024-06-13
Update installer.sh

MrTumnis:
Not adding a new line was causing a "directory does not exist" error because the line was continuos with the '#'


MrTumnis:
I'm not sure if signing the CLA was necessary for such a small change, but I figured it wouldn't hurt.

LDprg (approved):
Seems good.

oskardotglobal (approved):
Ok perfect, thank you

PR #121, opened 2024-07-01
Enhance VM Detection and Improve FreeRDP Command Handling

KernelGhost:
1. Improved detection of running VMs.
    - Replaced `virsh list` with `virsh list --state-running --name` to directly filter for running VMs by name.
    - Updated `grep` command to `grep -q '^RDPWindows$'` for precise matching of the VM name using a regular expression.
2. Convert FreeRDP command to array format for improved readability and straightforward modification.
3. Extract VM IPv4 address using a regular expression.

### DIFF
--- a/bin/winapps
+++ b/bin/winapps
@@ -65,8 +65,8 @@ if [ -z "$RDP_IP" ]; then
         echo "  sudo usermod -a -G kvm $(whoami)"
         exit
     fi
-    if virsh list | grep -vq RDPWindows; then
-        echo "RDPWindows is not running, run:"
+    if ! virsh list --state-running --name | grep -q '^RDPWindows$'; then
+        echo "RDPWindows is not running. Please run:"
         echo "  virsh start RDPWindows"
         exit
     fi
--- a/bin/winapps
+++ b/bin/winapps
@@ -90,17 +90,58 @@ if [[ -n "$RDP_FLAGS" ]]; then
     FREERDP_COMMAND="$FREERDP_COMMAND $RDP_FLAGS"
 fi
 
-if [ "$1" = "windows" ]; then
-    dprint "WINDOWS"
-    COMMAND=("${FREERDP_COMMAND}" "/d:${RDP_DOMAIN}" "/u:${RDP_USER}" "/p:${RDP_PASS}" "/scale:${RDP_SCALE}" "+dynamic-resolution" "+auto-reconnect" "+home-drive" "/wm-class:\"Microsoft Windows\"" "/v:${RDP_IP}")
-    "${COMMAND[@]}" 1>/dev/null 2>&1 &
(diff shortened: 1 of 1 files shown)


LDprg:
Great PR. However I am not sure about 2 point, because I don't see how/why this is better.

KernelGhost:
@LDprg It is recommended to use arrays for command construction, particularly when dealing with potentially complex or dynamically generated commands. In this context, command construction using arrays provides two main advantages:
    1. Arrays automatically handle quoting of arguments, ensuring that spaces and special characters in arguments are properly preserved. I was previously experiencing authentication failures when using a username that contained spaces prior to this adjustment.
    2. Arrays handle empty variables gracefully by skipping them, preventing the formation of double spaces. This was previously an (albeit small) issue when using variable interpolation directly (e.g. `$MULTI_FLAG`).

Although there are additional benefits related to preventing injection vulnerabilities, these are less relevant in this particular scenario.

For improved readability and clarity, the section of concern can be rewritten as follows:

```bash
COMMAND=(
    "${FREERDP_COMMAND}"                 # FreeRDP Command/Executable
    "/d:${RDP_DOMAIN}"                   # Remote Desktop Connection Domain
    "/u:${RDP_USER}"                     # Username
    "/p:${RDP_PASS}"                     # Password
    "/scale:${RDP_SCALE}"                # Remote Desktop Session Scaling Factor
    "+dynamic-resolution"                # Enable Dynamic Resolution Adjustment
    "+auto-reconnect"                    # Enable Automatic Reconnection on Disconnect
    "+home-drive"                        # Map User Home Drive
    "/wm-class:\"Microsoft Windows\""    # Window Manager
    "/v:${RDP_IP}"                       # Virtual Machine IP Address
)
```

LDprg (approved):
Seems good to me. @oskardotglobal can you also review?

bin/winapps
```diff
@@ -65,14 +66,12 @@ if [ -z "$RDP_IP" ]; then
         echo "  sudo usermod -a -G kvm $(whoami)"
         exit
     fi
-    if virsh list | grep -vq RDPWindows; then
-        echo "RDPWindows is not running, run:"
+    if ! virsh list --state-running --name | grep -q '^RDPWindows$'; then
```
oskardotglobal, line 69: Just a note: We should implement this check for docker sometime since virsh is deprecated.

oskardotglobal (approved):
Looks good to me.

bin/winapps
LDprg: @oskardotglobal We might want to check if there are other virsh depend checks.

KernelGhost: `virsh` is also used within the script to obtain the IPv4 address of the virtual machine. If `virsh` is deprecated, this may also need to be reconsidered.

PR #123, opened 2024-07-02
Reformat code, fix typos, use apt instead of apt-get, improve consistency

Kazevic:
Just some whitespaces were added to the SVG, no changes to the images themselves. Also, text that goes much past 120 characters is wrapped.


LDprg (approved):
Seems good to me.

PR #126, opened 2024-07-08
WinApps _requires_ FreeRDP version 3+, which is not clear from the README.md

bkanuka:
The flags used when calling `$FREERDP_COMMAND` are not compatible with FreeRDP v2.  Namely v2 uses `/app: ...` and `/app-cmd ...` instead of `/app:alias:...,cmd:...` syntax used in winapps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,14 +96,14 @@ If you already have a virtual machine or server you wish to use with WinApps, yo
 To get things going, use:
 
 ```bash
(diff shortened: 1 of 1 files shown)


LDprg (approved):
Thx that is much better!

oskardotglobal (approved):
LGTM.

PR #144, opened 2024-07-17
Fixed #141, #143 and #145

KernelGhost:
Modified `installer.sh` and `bin/winapps` to have distinct control flows for the deprecated `libvirt` and recommended `Docker` methods.

I believe removing support for the `libvirt`/`virt-manager` flavor of WinApps is reckless, especially since users may want the flexibility of using WinApps with an existing Windows VM. Therefore, instead of completely eliminating support for the older method, I chose to maintain both methods side-by-side.

This change ensures that dependency checks and calls to `virsh` are only run when the deprecated 'flavor' is specified within the WinApps configuration file. In addition, new dependency checks and calls to `docker` have been introduced. This includes a new function to check if the `Docker` container is powered on, resolving issues #141 and #145.

In addition, the logic used to determine the major FreeRDP version has been refined to resolve issue #143.

PR #148, opened 2024-07-17
Display warnings when changing RDP_SCALE automatically to prevent user confusion.

KernelGhost:
As requested by LDprg in PR #144.


LDprg (approved):
Seems good to me. Thx.

LDprg (approved):
@KernelGhost no problem, I am happy to see this fixes.

LDprg:
@oskardotglobal There is a difference between the two. 

Afaik Freerdp mounts the directory when connecting via the rdp protocol. Here qemu mounts the directory directly into the vm without the need of an connected client.

KernelGhost:
@oskardotglobal Yes, FreeRDP mounts the Linux home directory when `+home-drive` is used as an argument. As @LDprg mentioned, this additional mount allows access to the Linux home directory when the user is viewing the Windows VM via http://127.0.0.1:8006. I thought adding this additional mount might be useful during the initial setup of the VM, as the user may have certain files or programs stored on Linux that they wish to access from Windows prior to installing WinApps. This mount can be turned off by the user by commenting out the relevant line within `compose.yaml`.

KernelGhost:
My most recent set of commits add official support for `podman` as a daemonless alternative to using `docker` for the virtual machine backend. The relevant documentation has also been updated to reflect this.

KernelGhost:
The latest commit also fixes #91. Credit goes to [f8ith](https://github.com/f8ith) for identifying the cause of the issue.

KernelGhost:
I do not foresee making any further commits to this pull request. Please feel free to merge it when appropriate. Thanks!

oskardotglobal (commented):
LGTM.

PR #151, opened 2024-07-21
Added documentation for configuring `QEMU Guest Agent`

KernelGhost:
I was previously having issues where `virsh shutdown RDPWindows` would fail. Configuring `QEMU Guest Agent` allows the GNU/Linux host to properly shut down Windows instead of relying on ACPI commands. After completing the additional configuration specified in `docs/libvirt.md`, the aforementioned command now works.


oskardotglobal (approved):
Looks good to me.

PR #167, opened 2024-07-25
Adding more file type support for Microsoft Visio

gordoncheong:
Bit of an oversight from me in the last PR. This PR adds the mimetypes for Microsoft Visio templates and stencils along with their macro-enabled versions for a more complete experience.


oskardotglobal (approved):
LGTM.

PR #175, opened 2024-07-28
Added ability to automatically pause and resume Windows.

KernelGhost:
Added functionality to enable WinApps to automatically boot or resume Windows when a Windows application is launched, as well as pause Windows following a period of inactivity. This feature can be turned off by users via the WinApps configuration file.


LDprg (approved):
This was one of the features we talked about after the fork of the project. Great work @KernelGhost!

Code seems good to me. However, since this is a somewhat bigger feature I would prefer if this gets additionaly reviewed by @oskardotglobal too.

KernelGhost:
The original solution used a serial connection and a PowerShell script to communicate the state of the Windows VM to GNU/Linux, with a bash script listening for state changes and triggering a suspend following a period of inactivity. However, I ultimately opted for a simpler approach: Suspending Windows following an absence of `FreeRDP` processes previously initiated by WinApps. This method avoids the need for additional communication between host and guest. I'd appreciate it if Oscar could review this as well.

oskardotglobal (changes requested):
I'm pretty sure `docker compose pause` doesn't work with our specific docker image (I've recently tested it on my machine and it didn't work). We would probably have to look into pausing the VM by running shell commands from inside the container.

Could one of you test the pausing behaviour on podman?

installer.sh
```diff
@@ -82,6 +82,8 @@ RDP_PASS=\"MyWindowsPassword\"
 #RDP_FLAGS=\"\"
 #MULTIMON=\"true\"
 #DEBUG=\"true\"
+#AUTOPAUSE=\"on\" # Acceptable values are 'on' and 'off'.
+#AUTOPAUSE_TIME=\"300\" # Seconds before pausing Windows due to inactivity. Ignored if AUTOPAUSE 'off'.
```
oskardotglobal, line 86: You should probably add that this should be above 20 like mentioned in the source code above.

KernelGhost: The 20-second delay only applies to RDP sessions involving individual applications, not to full Windows RDP sessions. For full Windows RDP sessions, an explicit logoff signal is sent to the RDP server when the user logs off. I am wondering if there is a clearer way to explain this to users within the configuration file.

KernelGhost:
@oskardotglobal The command `podman-compose --file="${HOME}/.config/winapps/compose.yaml" pause` successfully pauses the Windows virtual machine on my system. Similarly, the `unpause` command also functions correctly with `podman`.

installer.sh
oskardotglobal: Well, we mostly do not use full RDP session in our case, so just saying it has to be a minimum of 20 seconds should suffice.

oskardotglobal:
Because I'm on vacation right now and won't be able to retest the behaviour on my machine for several weeks

KernelGhost:
@oskardotglobal I will re-install `Docker` and Windows and get back to you.

installer.sh
KernelGhost: @oskardotglobal Please see b90fb086a97893df2341d0f07ec1ffd4a778b8b0.

oskardotglobal (changes requested):
Autopause should be off by default and it should be mentioned that Docker is incompatible with it.
Alternatively, we make podman the default and keep autopause on by default.

KernelGhost:
The requested changes have been made in 6f439a4df9a4f869ddc8c415dcc49d4920ecc813.

oskardotglobal (approved):
LGTM.

LDprg:
> @oskardotglobal I've opened a new issue with the `dockur/windows` team [here](https://github.com/dockur/windows/issues/674). Is this the same error you encountered?
> 
> To be honest, I switched from using `docker compose` to `podman-compose` following problems with `Docker` failing to detect and utilise KVM, as well as due to the benefits of `podman` being rootless and daemonless. Since `podman-compose` is *supposed* to be fully compatible with `docker compose`, I didn't expect this to cause any issues.

That's the reason why I prefer podman over docker. Docker is sometimes a little bit bugged (especially when using a rootless install of docker).

PR #182, opened 2024-08-01
Auto install script for docker/dockur/winapps with 2 network options (DHCP LAN + macvlan & host network)

itiligent:
#### **Option 1**: [Easy] Deploys a Windows container with Dockur's default network settings 
  - A fully automated Docker/Dockur/Windows build with just a few simple prompts
  - The Windows container shares an IP with the host
     - (Optionally) restrict incoming RDP & VNC access to 127.0.0.1

#### 2. **Option 2**: [Advanced] Deploys a Windows container with macvlan networking 
   - Macvlans allow a container to act as a full LAN client with a separate DHCP IP & MAC address
      - The script walks the user through the required CIDR inputs & automatically builds all routing schema

(Note: Script will auto install flatpak version of FreeRDP3 if a distro version is not available)

### Script Instructions
 1. Run the script (without root permissions) & answer the script prompts to customize your build
 2. Follow the Windows install via VNC http://127.0.0.1:8006 or http://x.x.x.x:8006
 3. Restart Windows via the VNC console when the installation is complete
 4. Refresh Docker group permissions by running `newgrp docker`
 5. Test rdp from Linux by running `~/test-rdp.sh` (this test script is created at install to suit the individual build settings)
 6. Install your preferred Windows applications as needed via RDP or VNC
 7. Run  `~/winapps/installer.sh` to install WinApps

      
#### More about macvlans:
   - Macvlan pros:
     - Allows 3-way sharing between the container, local host, and LAN
     - Provides direct communication between the container and NIC, (no NAT or port forwarding)
     - DHCP makes IP management consistent, aiding firewall rules and network policy
     - Some applications may require unique IP/MAC addresses for licensing
     - (Optionally) restrict incoming RDP & VNC access to 127.0.0.1
     - Automates a complex build into a repeatable task with potential for rapid & low cost DR virtual desktop infrastructure 
   - Macvlan cons:
    - As macvlans use multiple MAC addresses per interface:
       - This option is available only for ethernet-connected systems
       - Macvlans are NOT compatible with WiFi NICs

The script is curently Debian/Ubuntu focused but can be used as a framework for other distros

### DIFF
--- a/winapps-for-dockur-setup.sh
+++ b/winapps-for-dockur-setup.sh
@@ -56,7 +56,7 @@
 ############################################################################################################
 
 # Container defaults
-DEFAULT_VERSION="tiny11" # see https://github.com/dockur/windows for all options
+DEFAULT_VERSION="win10" # see https://github.com/dockur/windows for all options
 DEFAULT_INSTALL_ISO="/home/david/w10_ent_iot_ltsc_2021_x64.iso"  # eg. /home/$USER/your-installer.iso
 DEFAULT_HOSTNAME="RDP_Windows"
 DEFAULT_USERNAME="user"
@@ -90,11 +90,11 @@ set -e
 
 echo
 printf "%b\n" "${GREEN}## WinApps for Dockur/Windows v1.0 ################################################"
-printf "%b\n" "${BLUE} 1. ${NC}Answer script prompts to suit your build"
-printf "%b\n" "${BLUE} 2. ${NC}Follow the Windows install via VNC ${BLUE}http://127.0.0.1:8006 or http://x.x.x.x:8006"
-printf "%b\n" "${BLUE} 3. ${NC}IMPORTANT! Restart Linux when the Windows installation is complete"
-printf "%b\n" "${BLUE} 4. ${NC}Install your preferred Windows applications via RDP or VNC"
-printf "%b\n" "${BLUE} 5. ${NC}Run ${BLUE}~/winapps/installer.sh${NC} to install WinApps"
+printf "%b\n" "${ORANGE} 1. ${NC}Answer script prompts to suit your build"
+printf "%b\n" "${ORANGE} 2. ${NC}Follow the Windows install via VNC ${BLUE}http://127.0.0.1:8006 or http://x.x.x.x:8006"
+printf "%b\n" "${ORANGE} 3. ${NC}IMPORTANT! Restart Linux when the Windows installation is complete"
+printf "%b\n" "${ORANGE} 4. ${NC}Install your preferred Windows applications via RDP or VNC"
+printf "%b\n" "${ORANGE} 5. ${NC}Run ${BLUE}~/winapps/installer.sh${NC} to install WinApps"
 
 # Make sure the script is NOT being run as root
 if [[ $EUID -eq 0 ]]; then
@@ -169,7 +169,7 @@ while true; do
                 else
                     echo "    Error: The default ISO file does not exist at ${DEFAULT_INSTALL_ISO}"
                     echo "    Please check the path and try again."
-                    echo
+					echo
                 fi
             else
                 echo "    No default ISO is set. Please enter 2 or 3."
@@ -178,13 +178,13 @@ while true; do
             ;;
     2)
         read -r -p "    Enter the path to your Windows install ISO: " custom_iso
-        if [ -e "$custom_iso" ]; then
+		if [ -e "$custom_iso" ]; then
                     INSTALL_ISO="${custom_iso}"
                     break
                 else
                     echo "    Error: The ISO file does not exist at ${custom_iso}"
                     echo "    Please check the path and try again."
-                    echo
+					echo
                 fi
         ;;
     3)
@@ -193,7 +193,7 @@ while true; do
         ;;
     *)
         echo "    Invalid choice. Please enter 1, 2, or 3."
-        echo
+		echo
         ;;
     esac
 done
@@ -231,7 +231,7 @@ while true; do
         break
     else
         echo "    Passwords do not match. Please try again."
-        echo
+		echo
     fi
 done
 
@@ -262,7 +262,7 @@ disable_sound() {
 
 while true; do
     read -r -p "    Enable sound? (yes/no) [default: ${DEFAULT_ENABLE_SOUND}]: " choice
-    choice="${choice:-${DEFAULT_ENABLE_SOUND}}"
+	choice="${choice:-${DEFAULT_ENABLE_SOUND}}"
 
     case "$choice" in
     yes | YES | y | Y)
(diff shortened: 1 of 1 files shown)


itiligent:
I made a final change to add a sound on/off prompt menu and simplified install instructions as a reboot seems best to avoid group permissions not always refreshing after docker is added via one continuous script.

oskardotglobal:
I don't know about this one. 
The script makes a lot of assumptions and there is no documentation for it. For this to be merged the according installation documentation would have to be updated (but only for Debian-based distros).

itiligent:
Firstly, this is a great project which is why I'd like to help it towards user friendliness & made more accessible for less technical folk. 

Regarding documentation, the script is a actually a coalescence of the very minimal amount of installation documentation I've been able to uncover. Another view (and a key part of my intent for submitting it) might be that reverse engineering it (its all  quite clearly commented) would help produce a much more comprehensive set of real world install instrutions than exists anywere because in writing it I uncovered many gaps in the current docs that had to be worked though.  

Could you please explain what documetation would be needed? Perhaps I could help?   At a minimum this scipt could perhpas be viewed as the quantum of actual verified install steps to build the WinApps recommended enviroment?

Regarding script assumptions, I dont quite understanfd what you mean there. There are virtually no assumprions contained (as far as I can tell) and the user is guided through every key decision point I've uncovered in my own use of WinApps.. Could you please give some examples of assumptions and perhaps I could address those? 

Regarding Debian based:   perhaps this is what you mean by assumptions?  I've put this install flow together as a kind of orderly inatall approach with a view that it could be expanded to other mainstream distros in the same way as I've smoothed out Docker's repo awkwardess od install between Debian distros.  

Happy to help!

winapps-for-dockur-setup.sh
```diff
@@ -0,0 +1,780 @@
...
+    ;;
+esac
+
+# Prompt for a custom OS install source
+echo
+while true; do
+    echo " Select a Windows installer source:"
+    echo "    1) Use your own ISO: ${DEFAULT_INSTALL_ISO}"
+    echo "    2) Provide a new ISO path"
+    echo "    3) No ISO (script will download Windows)"
```
oskardotglobal, line 148: This should be the default.

winapps-for-dockur-setup.sh
```diff
@@ -0,0 +1,780 @@
...
+      - ${NET_ACCESS_OPTION}8006:8006
+      - ${NET_ACCESS_OPTION}3389:3389/tcp
+      - ${NET_ACCESS_OPTION}3389:3389/udp
+    stop_grace_period: 120s
+    restart: on-failure
+    volumes:
+      - data:/storage
+      - ${HOME}:/shared
+      - ./oem:/oem
+      - $INSTALL_ISO:/custom.iso
```
oskardotglobal, line 679: Pretty sure this doesn't work even with your above empty image, just don't add the line if INSTALL_ISO is empty

winapps-for-dockur-setup.sh
```diff
@@ -0,0 +1,780 @@
...
+# Start/stop/enable/disable the container
+#    docker compose -f file.yaml up -d # start container
+#    docker stop <container-name>      # stop  container
+#    docker update --restart=no <container-name> # diasble container auto start at boot
+#    docker update --restart=on-failure <container-name> # enable container auto start at boot
+#    sudo systemctl start|stop|disable macvlan-br.service # toggle macvlan routing config
+############################################################################################################
+
+# Container defaults
+DEFAULT_VERSION="win10" # see https://github.com/dockur/windows for all options
```
oskardotglobal, line 59: This should be either `win11` or `tiny11`.

oskardotglobal (changes requested):
I've given the script another review and I missed that it clones winapps before. Just some minor things that need addressing.

Also, I would really prefer if you overwrote the default compose.yaml instead of writing a new one since that would preserve all the comments we added to it. We could also use env vars with fallback values inside the compose file so you could search and replace them.

The other major thing missing is that this script has to be mentioned within the README.

LDprg:
I feel like we shouldn't merge an debian only install script.

It should at least support debian, arch and fedora distros. This shouldn't be too much of a problem to implement and still avoid alot of confusion.

itiligent:
Thanks for the feedback. Take your time and play with it and get a feel for it and what I've tried to achieve through it. More eyes will always help find the things I mght not have thought of.

winapps-for-dockur-setup.sh
itiligent: No problem, thats in there just coz I actaully use a win10 distribution point as testing with downloads takes too long.

winapps-for-dockur-setup.sh
itiligent: it works on all flavours of debian. U22 & U24 dont like a blank entry there - at line 643 there is a workaround that adds a dummy file there if no ISO is configured. This was a far simpler and self contained soluton rather than adding a conditional override and extra malarkey to achieve in a script that may need to support extra distros - but I'm open to suggestions to improve my bash! 

The above works because if there is no *working* ISO image in the path that is specified, Dockur just defaults back to the DEFAULT_VERSION value and installs Windows from a download.  It get tricky when you are trying to straddle and support two independent pieces of software, so I'm trying to keep it as simple as possible.

winapps-for-dockur-setup.sh
itiligent: keeping the default compose in full is a good idea for continuity of docs. The main reason I cut it down was to make it easier to work with and review. I notice in the past few days the winapps.config deault file has been improved, so this should also be made to match.

An the matter of sound support, I would really prefer that the sound on/off hack on line 586 to not be needed, and instead all just be toggled on/off via  flags in the winapps.conf file as per the FreeRD3 docs suggest  - this would also make user sound permanent as this hack will get overwritten as soon as /upgrade/re-clone event takes place.

winapps-for-dockur-setup.sh
oskardotglobal: Yeah, that was what I was talking about

itiligent:
Ok so I've done a bunch of work on this today.  

- I've fixed the blank ISO issue, (it urns out that the custom ISO path must be specified below all volumes in the compose file else windows  fails to find an image after 2nd boot)
- Added lots of validation to the ISO inputs to check to make sure that an ISO file also exisits before the scipt will accept it
-  Default docker compose template files have been re-added, 
- New winapps.conf template is incorporated.
-  Added ability to override the auto IP adressing for Macvlan (offers flexibility for cases where several containers might run a single host or there are multiple NICS and a need for more than a  /30)  
- Lots of spell checks and grammar tidy ups
-  I've also changed the colours to match the WinApps logo. 
 
My preference is to gather all the feedback on this debian version, and once everyone is happy with it, then I can look to expanding it to Fedora and Arch

oskardotglobal:
Thanks for the great work, @itiligent. We'll properly review this once you're ready

itiligent:
Closing as I have progressed a much more refined install script version, and I will re-submit again soon.

PR #185, opened 2024-08-02
Add Support for Microsoft Project

gordoncheong:
This PR adds installation support of Microsoft Project to WinApps, including installation info and icons. README.md is also appended with Microsoft Project.

As for the `mimetype`, it is not so easy to add support for since Linux always thinks that `.mpp` is the Musepack Audio Format. If I am able to solve this problem, I will start another PR.


oskardotglobal (approved):
LGTM. Merging

PR #195, opened 2024-08-04
Add automatic freerdp local installation and docker/podman support

LDprg:
This is very much experimental and by far not ready!


oskardotglobal:
Unpacking the RPM is an interesting approach to grab binaries. However, we could also just redistribute FreeRDP binaries built inside Actions (or maybe even unpacked from RPM too) but that wouldn't require users installing rpm2cpio.

LDprg:
> Unpacking the RPM is an interesting approach to grab binaries. However, we could also just redistribute FreeRDP binaries built inside Actions (or maybe even unpacked from RPM too) but that wouldn't require users installing rpm2cpio.

I have considered this too, but I wasn't able to get a somewhat stable download link from the actions. The rpm package has a download link, which should be safe in terms of availability. (Fast changing links makes the app easily outdated just because it cannot download freerdp).

I also played around with flatpaks, since they might got a better way to decompress, but I failed to download the binaries without the usage of flatpak itself. Flatpak as dependency would make it incompatible with for example distrobox.

However I completely agree that rpm2cpio shouldn't be dependency (there shouldn't be any external dependencies at all), but it is not that much of a problem to change this to a better solution afterwards.

oskardotglobal:
We can create releases and grab the latest release using the github api to get stable download links.

oskardotglobal:
Actually, I just stumbled across the source for rpm2cpio: https://github.com/ruda/rpm2cpio/blob/master/rpm2cpio.py and it seems like quite the simple package to port to pure Rust as a crate and we could combine it with the [threecpio crate](https://crates.io/crates/threecpio) to achieve the same thing without any system requirements.

LDprg:
@oskardotglobal should I close this in favour of #308?

oskardotglobal:
I'd say once I'll push some more changes to my branch you can rebase your PR on it

oskardotglobal:
@LDprg I think you're good to rebase your PR, but make sure to check for xfreerdp3 being installed before just grabbing it from the RPM as e.g. on Nix that won't work anyways

LDprg:
> @LDprg I think you're good to rebase your PR, but make sure to check for xfreerdp3 being installed before just grabbing it from the RPM as e.g. on Nix that won't work anyways

Will look into this as soon as I have time.

winapps/src/freerdp.rs
```diff
@@ -1,21 +1,113 @@
...
+        fn get_freerdp() -> Command {
+            tokio::runtime::Builder::new_multi_thread()
+                .enable_all()
+                .build()
+                .unwrap()
+                .block_on(Freerdp::install_freerdp());
+            Command::new(get_data_dir().join("usr/bin/xfreerdp"))
+        }
+
+        async fn install_freerdp() {
```
AkechiShiro, line 22: Is there a particular reason why we're doing the installation here ? I think this won't work for NixOS for instance, maybe we should have it elsewhere

AkechiShiro: I think it's better to assume the user is able to install freerdp on his side or use some other script for installation than putting inside the new rewrite as for Nix/NixOS we'll already need to patch this else this won't work.

oskardotglobal: Regarding Nix, I've packed and re-packaged the rewrite twice already and the Command::new does not find the xfreerdp binary, even when I wrap the executable to prefix PATH...

LDprg: I am generally not happy with the way I implemented this.
But I haven't got any better solution.

AkechiShiro: @oskardotglobal : if you could share the package I could take a look.

@LDprg : That would help yes but I'm not sure we can provide the package for all distros in the Rust code, I mean this will take a lot of time to get right and maintain up to date.

oskardotglobal: @AkechiShiro I think you misunderstood what this is doing, this is not fedora-specific even though it downloads the RPM; the RPM is unpacked to extract the binary inside (there's still issues with this though)

Regarding the Nix package, the main problem is that Nixpkgs stable does not contain nightly rustc and I'd like to maintain compat with stable nix, so I'm currently using Flakes, see https://github.com/oskardotglobal/winapps/blob/feat-rewrite-api/flake.nix

AkechiShiro: Regarding rust-nightly, I think you'd need to use this : https://github.com/nix-community/fenix @oskardotglobal 

I also was able to use the packages without using Flakes, I think we should try to send them upstream in Nixpkgs and we can also let the users who want to use the Flake come to the repo and use it.

If the PR upstream to Nix gets denied, then we'll have tried to upstream it but most likely we'll rework on the rewrite to address any heavy issues or blocks for upstreaming the package in Nixpkgs.

LDprg: @AkechiShiro I did use fenix for projects before. It is a great way to use rust with nix.

oskardotglobal: @AkechiShiro I've used the Oxalica rust-overlay, which is based on Fenix; however, I don't think I would be able to use that in a Nixpkgs, would I?
The package definitely builds without flakes using flake-compat but that still requires us to write a flake.

AkechiShiro: Yeah it would not be usable in Nixpkgs I think, I'll have a look at some packages in Nixpkgs and check but I'd guess they override the rust package and bump the version post likely.

AkechiShiro: I think we'd be forced to override all rust related packages to use force use of nightly rust builds if this were to land in Nixpkgs

LDprg:
I feel like this should be redone completely. The implementation of this is currently bad, not ready at the very basics and alot changed in #308.

I am closing this in favour of a future PR. I will rethink the implementation (maybe snap or flatpak packages are better etc)!

PR #208, opened 2024-08-11
optimize the logic for detecting the FreeRDP command

escapefreeg:
Optimize the logic for detecting the FreeRDP command, falling back from `xfreerdp` to `xfreerdp3` to handle situations where both FreeRDP2 and FreeRDP3 are installed on the host


oskardotglobal (approved):
LGTM. Merging

Issue #216, opened 2024-08-15
Exiting with status '10'


soyposmo:
Hello. I am having trouble continuing the installation. I receive the message during the installation, both in the system and in the user. It stops and sends the following message:

```
################################################################################
#                                                                              #
#                            WinApps Install Wizard                            #
#                                                                              #
################################################################################

Q) Install or uninstall WinApps? --> Install
Q) Configure WinApps for the current user 'cosmo' or the whole system? --> System
Q) Automatically install supported applications or choose manually? --> Manual (Default)

[sudo] contraseña para cosmo: 
Installing WinApps.
Checking for existing conflicting WinApps installations... Done!
Attempting to load WinApps configuration file... Done!
Checking whether dependencies are installed... Done!
Checking whether the user 'cosmo' is part of the required groups... Done!
Checking the status of the Windows VM... Failed!

ERROR: WINDOWS VM DOES NOT EXIST.
The Windows VM 'RDPWindows' could not be found.
--------------------------------------------------------------------------------
Please ensure a Windows VM with the name 'RDPWindows' exists.
--------------------------------------------------------------------------------
Exiting with status '10'.

```
I have verified that the virtual machine is active and running, and that it has the correct name:
```
virsh --connect qemu:///system list --all             
 Id   Nombre       Estado
-------------------------------
 1    RDPWindows   ejecutando
```
I have added the path to my shell.
`LIBVIRT_DEFAULT_URI=qemu:///system`

This is how the configuration file of my virtual machine with Windows 10 looks
```
<domain type="kvm">
  <name>RDPWindows</name>
  <uuid>7a21ff2b-6943-4315-ade1-4ebb4d6a5134</uuid>
  <metadata>
    <libosinfo:libosinfo xmlns:libosinfo="http://libosinfo.org/xmlns/libvirt/domain/1.0">
      <libosinfo:os id="http://microsoft.com/win/10"/>
    </libosinfo:libosinfo>
  </metadata>
  <memory unit="KiB">4194304</memory>
  <currentMemory unit="KiB">1048576</currentMemory>
  <vcpu placement="static">2</vcpu>
  <os>
    <type arch="x86_64" machine="pc-q35-9.0">hvm</type>
    <boot dev="hd"/>
  </os>
  <features>
    <acpi/>
    <apic/>
    <hyperv mode="custom">
      <relaxed state="on"/>
      <vapic state="on"/>
      <spinlocks state="on" retries="8191"/>
      <vpindex state="on"/>
[...]

PR #218, opened 2024-08-18
Installer VM detection now not impacted by system language

### LINKED ISSUE #217: Installer is only supporting english system for VM Detection
Hello! First let me thank you for this awesome project !!! It's working perfectly in my laptop.

So I encountered this issue yesterday, and the same issue has been raised in #216 

Currently, the installer script will not work if the user has not an english system, because `virsh` will output english messages ("running", "shutoff" ...). And the error is not really easy to understand, as shown in the issue linked above.

the script is checking the vm status with harcoded  values:
https://github.com/winapps-org/winapps/blob/main/installer.sh#L787
https://github.com/winapps-org/winapps/blob/main/installer.sh#L749
A workaround is to run the installer with `LANG=C`

But we can make it language agnostic by using:
```console
virsh list --state-running
virsh list --state-paused
virsh list --state-shutoff
```

I will submit a pull-request :slightly_smiling_face:

Coruscant11:
Fixes #217 and #216 
Let me know if there is anything to improve!
I am not really used to shell scripts :smile:

### DIFF
--- a/installer.sh
+++ b/installer.sh
@@ -746,9 +746,14 @@ function waCheckVMRunning() {
     local VM_STATE="" # Stores the state of the Windows VM.
 
     # Obtain VM Status
-    VM_STATE=$(virsh list --all | grep -w "$VM_NAME")
+    VM_PAUSED=0
+    virsh list --state-paused | grep -wq "$VM_NAME" || VM_PAUSED="$?"
+    VM_RUNNING=0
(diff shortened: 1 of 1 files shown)


LDprg (approved):
Seems a better way to handle this stuff.

oskardotglobal (approved):
Looks good.

KernelGhost (approved):
LGTM

Issue #169, opened 2024-07-26
No Audio with /sound flag under WinApps, but running xrdp with /sound flag works fine


itiligent:
I am testing on two fresh systems of LMDE6 and LM 21 Ubuntu and here is what I am observing in both:

- Manually connecting to a Windows 10 Docker guest with xrdp shows the /sound flag working ok and an audio device is also present in the Windows mmsys. cpl. 
- However if I use /sound as an RDP_FLAG, sound does not work for apps launched by WinApps and there is also no audio device available in the mmsys. cpl if I run this from a WinApps cmd prompt.

I've spent hours on this trying many things sudh as /sound:sys:alsa, then installed Pulse and tried /sound:sys:pulse. I've also tried flatpak FreeRDP and freerdp3-x11 and all give the same results. For giggles I've also tested sound from the Windows container through a Guacamole jump host which also works ok, so I'm quite comfortable there's no issue with either Windows or the Linux host it would seem.

I cant see any other issues raised on this, perhaps there is something I am missing?

LDprg:
@itiligent  why would you want to set /sound? Winapps automatically sets /audio-mode:1 which i guess wont work combined with /sound and enables audio.

I am not sure what the goal is, since you didn't mentioned audio problems? Please clarify what problem you trying to solve.

itiligent:
I neglected to say (but in my own mind inferred) that with the default settings of RDP_FLAGS="" (unset) in the WinApps config sound does not work and no audio devices are presented to apps launched by WinApps. (The Fmstat legacy version used to work fine with an /sound:sys:alsa setting on the same hardware)
 
I tested with the /sound flag when manually running running xrdp. This info was provided to explain the observation that in all other aspects the underlying FreeRDP3 system on the Linux host appears to be working fine.   

The goal is to simply have sounds from Windows apps such as audible alerts for upcoming meetings and all other manner of sound based inteactions we are accustomed to responding to from windows like many other RDP clients out of the box.

acrilique:
I had to set audio-mode:0 and then I got sound working. Is there a consensus on whether we want the default to be audio on the remote (on the docker container, basically disabled in our use case from my perspective) which is what audio-mode:1 stands for, or audio-mode:0? If the general feeling is to have audio disabled by default, I'll make a PR to mention how to activate audio somewhere in the docs, or the readme, or a new md file. If not, we could change it to audio-mode:0 and also mention how to disable audio in the docs.

itiligent:
[...]

PR #219, opened 2024-08-19
Improve sound and microphone sharing.

RheaBarar:
Replaced `/audio-mode:1` with `/sound /microphone`. May help address #169.

### DIFF
--- a/bin/winapps
+++ b/bin/winapps
@@ -509,7 +509,8 @@ function waRunCommand() {
             +dynamic-resolution \
             +auto-reconnect \
             +home-drive \
-            /audio-mode:1 \
+			/sound \
+			/microphone \
             /wm-class:"Microsoft Windows" \
             /t:"Windows RDP Session [$RDP_IP]" \
             /v:"$RDP_IP" &>/dev/null &
@@ -528,7 +529,8 @@ function waRunCommand() {
             +auto-reconnect \
             +clipboard \
             +home-drive \
-            /audio-mode:1 \
+            /sound \
+			/microphone \
             -wallpaper \
             +dynamic-resolution \
             "$MULTI_FLAG" \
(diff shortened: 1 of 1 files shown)


KernelGhost:
This seems like a better way to do things. I have confirmed this change works well with my `libvirt`-based Windows virtual machine. LGTM.

KernelGhost (approved):
This seems like a better way to do things. I have confirmed this change works well with my libvirt-based Windows virtual machine. LGTM.

itiligent:
This does not solve #169
Would't the most intuitive way to solve this be to remove all reference to /sound and /microphone merged in b6f6a93 and instead simply pass these audio values to winapps following using the very standard convention a config file? i.e.   ~/.config/winapps/winapps.conf.

The way this is currently in implemented in b6f6a93 produces the same issue in #169 whereby sound config settings buried deep in git source *override* whatever audio values are added to the config file.  Also, any manual audio settings changes made to source will get overwritten with subsequent git clone updates or reinstalls, whereas a good old config file will keep the desired audio values regardless....which to me makes far more sense (and would have saved me a few wasted evenings a few weeks back!)

Or.. am I missing something as to why it must be done the current way?

KernelGhost:
@itiligent, you make a good point. There's no need for this to be handled in the shell script instead of the configuration file. Could you please submit a PR to move the audio settings `/sound` and `/microphone` into the default configuration file template?

oskardotglobal:
While we're at it, `/cert:tofu` should also be moved there

itiligent:
Ok, to clarifiy before I make a PR, you would like to remove instances of /cert:tofu, /sound & /microphone in the /bin/winapps file, and instead add these as (enabled) default options in the RDP_FLAGS line?  e.g. `RDP_FLAGS="/cert:tofu /sound /microphone"`

KernelGhost:
That looks perfect, yes. Could you also please test the changes before submitting the PR? Thanks so much!

PR #232, opened 2024-08-28
Simplify installation process

### LINKED ISSUE #213: "Please ensure Windows is powered on" wrong help text
I mistakenly used the compose.yml file as described by the dokur/windows project, and thus didn't have the correct container name set to `WinApps`, which makes the install script fail with the following message:

```
ERROR: CONTAINER NOT RUNNING.
Windows is not running.
--------------------------------------------------------------------------------
Please ensure Windows is powered on:
podman-compose --file ~/.config/winapps/winapps.conf start
--------------------------------------------------------------------------------
```

Clearly it is wrongly using the `winapps.conf` filename in the compose command. also if the container is not running, chances are it should be created with `up -d` instead of just a `start` command, at least as long as 'stoped' is not returned by the container status a few lines above:

```
CONTAINER_STATE=$("$WAFLAVOR" ps --all --filter name="WinApps" --format '{{.Status}}')
```

oskardotglobal:
This PR does the following:
- `installer.sh` is renamed to `setup.sh`
- The `setup.sh` and `bin/winapps` are now symlinked to `$BIN_PATH`
- It's checked that `$BIN_PATH` is on `$PATH`
- The script also now clones the repo to a default location in `~/.local/bin/winapps`

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,33 +284,25 @@ The following guides are available:
 
 If you already have a Windows VM or server you wish to use with WinApps, you will need to merge `install/RDPApps.reg` into the Windows Registry manually.
 
-### Step 2: Clone WinApps Repository and Dependencies
-1. Clone the WinApps GitHub repository.
-    ```bash
-    git clone --recurse-submodules --remote-submodules https://github.com/winapps-org/winapps.git && cd winapps
-    ```
-
-2. Install the required dependencies.
-    - Debian/Ubuntu:
-        ```bash
-        sudo apt install -y dialog freerdp3-x11 iproute2 libnotify-bin netcat
-        ```
-    - Fedora/RHEL:
-        ```bash
-        sudo dnf install -y dialog freerdp iproute libnotify nmap-ncat
-        ```
-    - Arch Linux:
-        ```bash
-        sudo pacman -Syu --needed -y dialog freerdp iproute2 libnotify gnu-netcat
-        ```
-    - Gentoo Linux:
-        ```bash
-        sudo emerge --ask=n sys-libs/dialog net-misc/freerdp:3 net-misc/iproute2 x11-libs/libnotify net-analyzer/netcat
-        ```
-
-- `Docker Engine`: https://docs.docker.com/engine/install/
-- `Podman`: https://podman.io/docs/installation
-- `podman-compose`: https://github.com/containers/podman-compose
+### Step 2: Install Dependencies
+Install the required dependencies.
+  - Debian/Ubuntu:
+      ```bash
+      sudo apt install -y dialog freerdp3-x11 iproute2 libnotify-bin netcat
+      ```
+  - Fedora/RHEL:
+      ```bash
+      sudo dnf install -y dialog freerdp iproute libnotify nmap-ncat
+      ```
+  - Arch Linux:
+      ```bash
+      sudo pacman -Syu --needed -y dialog freerdp iproute2 libnotify
+      gnu-netcat
+      ```
+  - Gentoo Linux:
+      ```bash
+      sudo emerge --ask=n sys-libs/dialog net-misc/freerdp:3 net-misc/iproute2 x11-libs/libnotify net-analyzer/netcat
+      ```
 
 > [!NOTE]
 > WinApps requires `FreeRDP` version 3 or later. If not available for your distribution through your package manager, you can install the [Flatpak](https://flathub.org/apps/com.freerdp.FreeRDP).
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -284,33 +284,25 @@ The following guides are available:
...
-    - Arch Linux:
-        ```bash
-        sudo pacman -Syu --needed -y dialog freerdp iproute2 libnotify gnu-netcat
-        ```
-    - Gentoo Linux:
-        ```bash
-        sudo emerge --ask=n sys-libs/dialog net-misc/freerdp:3 net-misc/iproute2 x11-libs/libnotify net-analyzer/netcat
-        ```
-
-- `Docker Engine`: https://docs.docker.com/engine/install/
```
LDprg, line 311: I think we should keep the references for docker and podman.

LDprg (commented):
Great, this would simplify alot and prevent a lot of problems related to local git repos.

README.md
oskardotglobal: I removed it because it makes it seem like you have to install all of the dependencies and it's also redundant since it's mentioned in the VM creation guide anyways

LDprg (approved):

KernelGhost (approved):
These are some fantastic and welcomed changes. Unfortunately my computer is out of order currently so I have not been able to test these changes. Based on a cursory inspection, LGTM.

oskardotglobal:
@LDprg Could you test this on your system?

LDprg:
> @LDprg Could you test this on your system?

I will take a look soon, but I am currently very low on time.

AkechiShiro:
On which Distro would you like me to test this @oskardotglobal (Debian or ArchLinux ?) looking to move forward #234

oskardotglobal:
I also run Nix and I've tested it there, but if you test it on another distro try Debian

AkechiShiro:
Ran into this issue after installing deps (curl and dialog) @oskardotglobal 
image

AkechiShiro:
I think the issue comes from here : 
image

AkechiShiro:
Removing `readonly` doesn't seem to fix it : 
image

oskardotglobal:
@AkechiShiro This should fix your problem

AkechiShiro:
Seem to have an issue again @oskardotglobal 

I don't think you're checking if the current user is in `sudoers` group also what if user is using something else than `sudo` like `doas` ?

I'm just trying it as a total Linux newbie to make sure we squash easy stuff early on.

image

AkechiShiro:
Didn't follow the steps properly but I think we should create the config file by default instead of asking the user to do another step : 
image

AkechiShiro:
I think dependency checks should be done at the beginning, I've had to install 2 deps before starting the script and now it is asking for more : 
image

I can help on reworking the PR to make it better if needed

AkechiShiro:
This dependency does not exist on Debian : 
image

oskardotglobal:
These issues are not related to this PR, but need addressing anyways

AkechiShiro:
Seems freerdp3 is not available on Debian...

I probably need to enable backports from unstable branches : https://packages.debian.org/bookworm-backports/freerdp3-x11

oskardotglobal:
Either that or use Flatpak, but that's kind of broken
We need a solution for this, ideally we'd be able to ship FreeRDP with winapps

oskardotglobal:
Or you know, just tell everyone to install the Nix package manager and install WinApps through there

AkechiShiro:
We both know that's not gonna work out XD, but someday maybe Nix will become mainstream

AkechiShiro:
We can ship Nix with Winapps and have it pull freerdp3

AkechiShiro:
We can use @oskardotglobal : https://github.com/matthewbauer/nix-bundle

oskardotglobal:
> We both know that's not gonna work out XD, but someday maybe Nix will become mainstream

We can always dream ;D

Nix-bundle seems like a good idea, we can experiment with that

oskardotglobal:
Actually, it seems like not that straight forward. It seems we'd need our package to be on nixpkgs or at least inside a nixpkgs fork and our package emits 2 binaries whilst nix-bundle seems to only emit one. This might be fixable if we split installer and winapps itself into two nix packages though

oskardotglobal:
I think this PR can be merged now though, we can continue this discussion in #234.

AkechiShiro:
I thought I'd have time to do a retest before the merge, so we can identify potential improvement points and mark them as open issues, that would get resolved in further issues.

oskardotglobal:
I did test this extensively, but the bugs you are facing are unrelated to this PR itself; they are general problems with how the installer works.

AkechiShiro:
Then we should open issue to track them, because I don't know them by heart on my side and I will forget about them after a few weeks

PR #237, opened 2024-09-01
README.md: Rename KDE Plasma shorthand to the correct `Plasma`

C0rn3j:
Shorthand for KDE Plasma is Plasma, not KDE.  
KDE is the group, Plasma is the DE.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 <p align="center"><img align="center" width="700" src="./icons/banner_dark.svg#gh-light-mode-only"/></p>
 <hr>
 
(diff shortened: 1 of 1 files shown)


KernelGhost (changes requested):
This seems somewhat overly focused on semantics. KDE stands for "K Desktop Environment". In common usage, it's rarely referred to simply as "Plasma." It's more often called "KDE" or "KDE Plasma." It might be a good idea to update the documentation to use "KDE Plasma" instead of just "Plasma" for clarity.

C0rn3j:
> KDE stands for "K Desktop Environment".

> It's more often called "KDE" 

Last time this was true was in 2008, KDE is simply KDE for more than 15 years now, and the correct shorthand is Plasma.

https://kde.org/cs/plasma-desktop/

PR #15, opened 2023-07-15
Relicense legacy to Unlicense

oskardotglobal:
I have created a CLA, added it to CLA-Assistant, included the Code Covenant Code of Conduct and have written some contribution guidelines. We now only require fmstrat to sign the CLA and then we're golden.

PR #240, opened 2024-09-05
Resolve licensing issues

oskardotglobal:
Moving this here since Github didn't allow properly reverting #15.


oskardotglobal:
cc @GreatNovaDragon

oskardotglobal:
So @AkechiShiro suggested this license: https://commission.europa.eu/content/european-union-public-licence_en ([full license text](https://joinup.ec.europa.eu/sites/default/files/custom-page/attachment/eupl_v1.2_en.pdf)). 

This license is OSI-approved. However, [the FSF classifies it as non-compatible with the GPL](https://www.gnu.org/licenses/license-list.html#EUPL-1.2) because it technically allows sublicensing and relicensing the work under a GPLv3. In our case, this would not be possible due to the CLA, but I don't see the advantages of the EUPL right now.

AkechiShiro:
Could we then move from GPLv3 to AGPLv3 ? @oskardotglobal

oskardotglobal:
@AkechiShiro The only difference between the AGPLv3 and the GPLv3 is the source distribution clause when distributing it over a network. Since WinApps is distributed as it's source, I don't see why this would matter.

AkechiShiro:
Hmm I thought it would be useful if someone tries to make a payed service with winapps without redistributing the source, nor providing the license we use.

https://opensource.stackexchange.com/questions/5041/of-the-differences-between-the-gplv3-and-the-agplv3-texts-what-to-make-of-them#5054

oskardotglobal:
Well, the changed Section 13 explicitly talks about a server, I don't see how one would make a paid service out of WinApps.

oskardotglobal:
Ok, this is now ready for review

AkechiShiro:
@oskardotglobal So you don't see how it would be possible to serve Winapps through an xrdp server or KasmVNC ?

oskardotglobal:
Well then you would be providing Windows VMs as a service, not WinApps, would you not?
We can't really stop someone from just selling VMs with RDP enabled and then "recommending" their users to use WinApps. But you're probably right, I can change this to an AGPLv3 if you want.

AkechiShiro:
I would like if there is no objection from anyone that we'd use AGPLv3 @GreatNovaDragon

GreatNovaDragon:
sure, agpl works for me

oskardotglobal:
Can someone review?

GreatNovaDragon (approved):

AkechiShiro:
@oskardotglobal can you rebase on main I see changes on CI and the setup script, this shouldn't be here.

oskardotglobal:
The CI changes are just to ensure the license headers on all files.

.pre-commit-config.yaml
```diff
@@ -48,3 +49,8 @@ repos:
     rev: v0.10.0.1
     hooks:
       - id: shellcheck
+
+  - repo: https://github.com/emzeat/mz-lictools
```
AkechiShiro, line 53: Could theses repos being used documented in case they get taken over by a malicious actor would be great to know what the CI depends on, without having to browse all the CI code.

CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,133 @@
...
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+[INSERT CONTACT METHOD].
```
AkechiShiro, line 64: We should have some mail or matrix channel here for moderation

apps/access-o365-x86/info
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2024 winapps-org
+# All rights reserved.
+#
+# SPDX-License-Identifier: Unlicense
```
AkechiShiro, line 4: This is alright ? You'll have to run the CI once to change this for all the files ??

apps/access-o365/info
```diff
@@ -1,3 +1,8 @@
+# Copyright (c) 2024 Jon Champagne
+# All rights reserved.
+#
+# SPDX-License-Identifier: Proprietary
```
AkechiShiro, line 4: I'm not sure why this is set to proprietary here

AkechiShiro (approved):

apps/access-o365/info
AkechiShiro: I think I understand that some files were contributed by others, which have set/used the license they wished to used before, thus the proprietary or other types showing up there.

oskardotglobal: Those are filed from the original repository, not by one of our contributors
No CLA means no license for us

.pre-commit-config.yaml
oskardotglobal: I don't really think a supply chain attack like this is very probable but we can fork the smaller repositories if you like

CODE_OF_CONDUCT.md
oskardotglobal: Yeah, although we should probably use one on `matrix.org` or some other instance, the federation of my homeserver keeps breaking so using that isn't a good idea

.pre-commit-config.yaml
AkechiShiro: No need to fork them but I just want to know exactly which one we are using so we can check with a quick glance in some documentation

.license-tools-config.json
```diff
@@ -0,0 +1,70 @@
...
+    ".bat": "BATCH_STYLE",
+    "": "POUND_STYLE"
+  },
+  "exclude": [
+    "^\\.[^/]+",
+    "/\\.[^/]+",
+    "^(.+)\\.(md|svg|png|reg|gif)",
+    "^fonts(/.*)?",
+    "apps/access/info",
+    "apps/access-o365-x86/info",
```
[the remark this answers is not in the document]
AkechiShiro: What do you mean ? Needs to be removed ???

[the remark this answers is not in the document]
oskardotglobal: Yes, it was a note for myself, I fixed that though

AkechiShiro:
I've created this matrix channel : https://matrix.to/#/#winapps:matrix.org feel free to join @oskardotglobal @LDprg @GreatNovaDragon

CODE_OF_CONDUCT.md
AkechiShiro: @oskardotglobal could you add the matrix channel to the contact method : [https://matrix.to/#/#winapps:matrix.org](https://matrix.to/#/#winapps:matrix.org)

.pre-commit-config.yaml
oskardotglobal: What do you mean by this? They are listed in the pre-commit config

AkechiShiro:
What is missing to merge this PR ?

.pre-commit-config.yaml
AkechiShiro: I just wanted them in the README I didn't check the precommit hook but I will thanks

oskardotglobal:
Nothing that I know of

AkechiShiro:
Maybe I'll open an documentation issue for the CI actions used, but I dont think there is anything else of concern

AkechiShiro:
Maybe I'll open an documentation issue for the CI actions used, but I dont think there is anything else of concern

AkechiShiro:
Hey @oskardotglobal and @LDprg could you join the Matrix/Element chat please ?

I would like to discuss about wayland support plans and I'm looking into this : https://github.com/elliotkillick/qvm-create-windows-qube

oskardotglobal:
Hold on I'll have to unfuck my homeserver's federation first

PR #245, opened 2024-09-08
Fixing Uninstall Failure

gordoncheong:
As of the latest git pull, the uninstall option fails to complete with the following error:

image

This is because the setup script was trying to remove the `winapps` directory on `$BIN_PATH` as a file in line 1676:
`$SUDO rm -f "${BIN_PATH}/winapps"`

Adding the recursive `-r` flag fixes this issue and uninstalls are successful.


oskardotglobal:
The bug you are facing has a different reason. Please try the fix I just committed.

gordoncheong:
Just tested it and the fix works great. Thanks for the edit.

oskardotglobal (approved):
Alright, then I can merge this.

PR #247, opened 2024-09-09
Added some old apps.

queler:
Added acrobat 9, Word and Excel 2010, and Microsoft Paint.
Acrobat in particular did not work with autodetect.


gordoncheong:
I think it's best to also update the Supported Applications List in one go, or else we will lose track of them eventually.

oskardotglobal (approved):
LGTM.

LDprg (approved):

PR #248, opened 2024-09-09
Support old windows versions

queler:
Added a test to not error out when the Appx cmdlets are not available (like in the windows 7 box i am using). I had some install issues with creating the bin links so I added "-f" and the absolute path.


LDprg (approved):
Nice little addition. Thx

oskardotglobal (approved):
LGTM.

KernelGhost (approved):
Looks good!

PR #250, opened 2024-09-10
switch managment of /sound /microphone & /cert:tofu to the winapps config file

itiligent:
As requested (and tested on debian 12). 
The default config file now the place to configure sound, mic and tofu. The default config file in readme now reads:
 `RDP_FLAGS="/cert:tofu /sound /microphone"`

I have also harmonised the common RDP options that apply when $FREERDP_COMMAND is called. The existing script was a little inconsistent across each of the 4 ways it could be called. Now the defauts that apply uniformly are:
```
+dynamic-resolution 
+auto-reconnect 
+home-drive 
+clipboard 
-wallpaper 
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -365,9 +365,9 @@ WAFLAVOR="docker"
 RDP_SCALE="100"
 
 # [ADDITIONAL FREERDP FLAGS & ARGUMENTS]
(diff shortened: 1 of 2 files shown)


oskardotglobal (approved):
Perfect, thank you.

PR #263, opened 2024-09-19
Fix netcat package typo for arch systems

MopigamesYT:
2. Arch Linux:

- Minor change: Consolidated the netcat package installation into a single line


freechelmi (approved):
LGTM

PR #264, opened 2024-09-19
Remove Winapps install from Docker doc

freechelmi:
This winapps paragraph leads user to confusion and start winapps right away although they have other steps to do in the main doc page

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -49,15 +49,6 @@ docker compose --file ./compose.yaml down
 docker compose --file ~/.config/winapps/compose.yaml up
 ```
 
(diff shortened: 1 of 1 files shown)


oskardotglobal (approved):
Good idea.

AkechiShiro:
Shouldn't it be moved and the instruction to setup be first linked there before this step ?

freechelmi:
> maybe we should put those instruction in a seperate doc `installation.md` to have a less longer README

Yes you can propose a modification to the readme but I'd rather spend time implementing the needed steps in the script

PR #268, opened 2024-09-23
Fix keyboard-layout being changed to US on RDP connection

tristanRW:
Added the IgnoreRemoteKeyboardLayout Registry Key to prevent automatic changes to the keyboard layout.

As mentioned in [this article](https://poweradm.com/ignoreremotekeyboardlayout-windows-rdp/) the rdp server adds and activates the US keyboard-layout if the keyboard-layout of the RDP client does not match the RDP servers.
Since our client is on linux it is very likely that this is the case.
This changes makes the RDP-session ignore the clients keyboard-layout and keeps the one that is configured on the server (= the one set up during the setup of the windows VM).


LDprg:
@oskardotglobal @KernelGhost What do you think? I think this could solve a lot of this weird keyboard issues.

oskardotglobal (approved):
LGTM.

PR #279, opened 2024-10-02
Fix timezone issues in Docker VM

eylenburg:
This registry key fixed an issue with the VM's timezone changing itself after each reboot. Related issue: https://github.com/dockur/windows/issues/672

### DIFF
--- a/oem/RDPApps.reg
+++ b/oem/RDPApps.reg
@@ -8,3 +8,6 @@ Windows Registry Editor Version 5.00
 
     [HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows NT\CurrentVersion\Winlogon]
     "AutoAdminLogon"="0"
+
+    [HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\TimeZoneInformation]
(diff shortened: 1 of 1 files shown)


itiligent:
This issue relates to docker and (I assume) podman containers using UTC by defaultt, which is a pain.  I've also been adding this suggested  reg key to RDPapps.reg file and it does indeed fix the local time issue on the next boot after install

KernelGhost (approved):
I don't personally use Docker or Podman to run WinApps, but based on the discussion in the linked issue and @itiligent's endorsement of this fix, I am happy to endorse merging this change.

itiligent:
A few extra thoughts.... while his change will make windows assume the docker container time is UTC. If using a qemu vm or similar, this might have unintended consequences unless it's documented or only deployed for container instances. Although  production hypervisors typically require system time is set to utc,  on the desktop with qemu this might not be as practical.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

oskardotglobal:
I think it would be better if this was only applied when in a container.

oskardotglobal:
Closing this in favour of #312.

PR #281, opened 2024-10-05
use netcat instead of netcat-gnu

CHN-beta:
On my machine, `nc` from netcat-gnu does not work, which makes winapps not work in manual mode (winapps thinks port 3389 not open, but it is actually open).

use `nc` from netcat works.


oskardotglobal (approved):
Alright, LGTM.

Thanks for contributing!

PR #291, opened 2024-10-08
README.md: Update Gentoo dependencies list

Username404-59:
Fixes the ebuild parent dirs & depend on openbsd-netcat instead of netcat because libvirt needs it and is likely to be installed


oskardotglobal (approved):
LGTM. Merging.

Issue #334, opened 2024-11-24
Windows session not starting


Boyfinn:
A couple days ago my WinApps setup stopped working. I've tried terminating and restarting the container, but when launching windows for example: I just get a toast saying: "WinApps: Booting Windows", and then nothing happens.
Trying to launch windows through the terminal gives no output either.

I'm on:
OS: Linux mint 22_x86_64
Kernel:  6.8.0-49
Desktop: Cinnamon 6.2.9

mindset-tk:
I'm also having this issue. Winapps.log is not much help.  Related issue that will confuse things, I can run a windows RDP session, and it runs fine. Below, the log states that windows immediately ended when in fact I ran the session for about 5 minutes, clicked around and opened terminal to test that it worked inside of the freerdp session. The "END" line was not appended until after I closed the session.

`[Mon Nov 25 11:52:37 AM PST 2024-21660] START
[Mon Nov 25 11:52:37 AM PST 2024-21660] SCRIPT_DIR: /home/elle/.local/bin
[Mon Nov 25 11:52:37 AM PST 2024-21660] SCRIPT_ARGS: WindowsTerminal
[Mon Nov 25 11:52:37 AM PST 2024-21660] HOME_DIR: /home/elle
[Mon Nov 25 11:52:37 AM PST 2024-21660] LAST_RUN: 1732564267
[Mon Nov 25 11:52:37 AM PST 2024-21660] THIS_RUN: 1732564357
[Mon Nov 25 11:52:37 AM PST 2024-21660] Using FreeRDP command 'xfreerdp3'.
[Mon Nov 25 11:52:37 AM PST 2024-21660] END
[Mon Nov 25 11:52:46 AM PST 2024-13636] START
[Mon Nov 25 11:52:46 AM PST 2024-13636] SCRIPT_DIR: /home/elle/.local/bin
[Mon Nov 25 11:52:46 AM PST 2024-13636] SCRIPT_ARGS: windows
[Mon Nov 25 11:52:46 AM PST 2024-13636] HOME_DIR: /home/elle
[Mon Nov 25 11:52:46 AM PST 2024-13636] LAST_RUN: 1732564357
[Mon Nov 25 11:52:46 AM PST 2024-13636] THIS_RUN: 1732564366
[Mon Nov 25 11:52:46 AM PST 2024-13636] Using FreeRDP command 'xfreerdp3'.
[Mon Nov 25 11:52:46 AM PST 2024-13636] WINDOWS
[...]

PR #336, opened 2024-11-25
Fix logging to show correct date/times

mindset-tk:
Logging was misleading and the date/times logged to file were always the time that the app was initialized, instead of the time of the actual log entry. Issue was noted in https://github.com/winapps-org/winapps/issues/334.

### DIFF
--- a/bin/winapps
+++ b/bin/winapps
@@ -42,7 +42,7 @@ readonly CONTAINER_NAME="WinApps" # FOR 'docker' AND 'podman' ONLY
 readonly RDP_PORT=3389
 readonly DOCKER_IP="127.0.0.1"
 # shellcheck disable=SC2155 # Silence warnings regarding masking return values through simultaneous declaration and assignment.
-readonly RUN="$(date)-${RANDOM}"
+readonly RUNID="${RANDOM}" #Set a random int for the run ID to assist with logging
 
 ### GLOBAL VARIABLES ###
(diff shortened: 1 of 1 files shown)


bin/winapps
```diff
@@ -162,7 +162,7 @@ Please run:
 # Name: 'dprint'
 # Role: Conditionally print debug messages to a log file, creating it if it does not exist.
 function dprint() {
-    [ "$DEBUG" = "true" ] && echo "[$RUN] $1" >>"$LOG_PATH"
+    [ "$DEBUG" = "true" ] && echo "[$(date)-$RUNID] $1" >>"$LOG_PATH"
```
oskardotglobal, line 165: The `RUNID` variable doesn't seem to be accessed anywhere else, so you can remove it.

oskardotglobal (changes requested):

LDprg (approved):

bin/winapps
mindset-tk: The run ID will aid in troubleshooting since it adds a semi unique random int to the end of the date on the log file, that remains the same across the each run of the script.  It would need to be stored in a var somewhere.

This was the apparent original purpose of declaring the $RUN var at the top of the script.

oskardotglobal: Ok, that seems logical.

oskardotglobal (approved):

Issue #350, opened 2024-12-07
More robust installation


oskardotglobal:
If the first run of `setup.sh` fails, the bootstrapping enters a broken state which can only be resolved by deleting `~/.local/bin/winapps-src`.

We should also look into making the installation simpler, the Nix package is already a step in the right direction.

PR #374, opened 2025-01-03
Fixup the bug that when the winapps-src is cloned but $INQUIRER is …

stceum:
…not updated with cloned path causing the failure of running the script.

Fixup #350


oskardotglobal (approved):
Looks good, thank you for your contribution

PR #387, opened 2025-01-15
waEnsureOnPathFix

starbr3aker:
If the $BIN_PATH is entirely missing from $PATH, throw error.


oskardotglobal (approved):
Can confirm this works. Thank you for your contribution.

PR #388, opened 2025-01-16
Add the reMarkable desktop app

toastedcrumpets:
I've added support for the reMarkable's desktop app
https://remarkable.com/
This is not "autodetected" by setup.sh in the unsupported application part, so adding it as a "supported" app enables it to be used.
I drew the icon for the app myself, its inspired by the normal icon design but I deliberately used a standard color and serif font to avoid being a direct replica, which is hard given how trivial the original design is.

Please let me know if you need anything else, and thank you for such an awesome tool!

### DIFF
--- a/apps/remarkable-desktop/info
+++ b/apps/remarkable-desktop/info
@@ -1,0 +1,22 @@
+# Copyright (c) 2025 Marcus Bannerman
+# All rights reserved.
+#
+# SPDX-License-Identifier: Proprietary
+
+# GNOME shortcut name
+NAME="reMarkable"
+
+# Used for descriptions and window class
+FULL_NAME="reMarkable Desktop App"
+
+# The executable inside windows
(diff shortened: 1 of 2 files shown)


apps/remarkable-desktop/info
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2025 Marcus Bannerman
```
oskardotglobal, line 1: Please drop this header so your contribution can be licensed under the AGPL. It is only included in some files because the authors did not sign the CLA.

oskardotglobal (changes requested):

apps/remarkable-desktop/info
toastedcrumpets: Sure, I just copied the word app, its removed now.

oskardotglobal (approved):
LGTM.

PR #402, opened 2025-02-02
Documentation Improvements

KernelGhost:
This PR features a series of improvements to the project documentation to address issues I encountered whilst installing WinApps on a new Ubuntu 24.10 system.

PR #405, opened 2025-02-02
fix(packages/winapps-launcher): Rename `WinAppsLauncher` related to `WinApps-Launcher`

mbekkomo:
The upstream has renamed the file `WinAppsLauncher.sh` to `WinApps-Launcher.sh`.


oskardotglobal:
I have gone ahead and addressed this as part of #402.

PR #427, opened 2025-02-16
Specify that iptables kernel module must be loaded for folder sharing…

raffaem:
…  to work

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -14,7 +14,7 @@ Although WinApps supports using `QEMU+KVM+libvirt` as a backend for running Wind
 > [!IMPORTANT]
 > The iptables kernel module must be loaded for folder sharing with the host to work.
 > Check that the output of `lsmod | grep ip_tables` and `lsmod | grep iptable_nat` is non empty.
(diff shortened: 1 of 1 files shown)


docs/docker.md
```diff
@@ -11,7 +11,10 @@ Although WinApps supports using `QEMU+KVM+libvirt` as a backend for running Wind
 > - It is recommended to edit the initial `compose.yaml` file to keep your required username and password from the beginning.
 > - It is recommended to not use `sudo` to force commands to run. Add your user to the relevant permissions group wherever possible.
 
-
+> [!IMPORTANT]
+> The iptables kernel module must be loaded for folder sharing with the host to work.
+> Check that the output of `lsmod | grep ip_tables` and `lsmod | grep iptable_nat` is non empty.
+> If it is, run `echo -e "ip_tables\niptable_nat" | sudo tee /etc/modules-load.d/iptables.conf` and reboot.
```
oskardotglobal, line 17: It is unclear whether to run this if the output of the command is empty or nonempty, it would be better to write "If it is empty"

oskardotglobal (changes requested):

oskardotglobal:
Thank you for your contribution. Just a small nit and then this can be merged.

oskardotglobal (approved):

PR #455, opened 2025-03-10
Adding Photoshop 2022 in pre-defined apps

limemane:


LDprg (commented):
Seems good to me.

LDprg (approved):

apps/photoshop-2022/info
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2024 Fmstrat
+# All rights reserved.
+#
```
oskardotglobal, line 3: Please remove this license comment

oskardotglobal (changes requested):

apps/photoshop-2022/info
limemane: ah yes, that makes sense. It is now removed

oskardotglobal (approved):

PR #475, opened 2025-04-02
fix: Change default to Win11 Pro

kroese:
The compose file currently defaults to Tiny11, but this is not ideal for two reasons:

  - There is only one download mirror for Tiny11 (archive.org, which is extremely slow).
  - It crashes on some systems, see: https://github.com/dockur/windows/issues/1177

The Win11 Pro preset has multiple high-speed download mirrors, and it doesnt crash on systems where Tiny11 does.

If people insist on using Tiny11, they can still do so by modifying `VERSION` or providing a .iso file, at their own risk.

### DIFF
--- a/compose.yaml
+++ b/compose.yaml
@@ -22,11 +22,12 @@ services:
       USERNAME: "MyWindowsUser" # Edit here to set a custom Windows username. The default is 'MyWindowsUser'.
       PASSWORD: "MyWindowsPassword" # Edit here to set a password for the Windows user. The default is 'MyWindowsPassword'.
       HOME: "${HOME}" # Set path to Linux user home folder.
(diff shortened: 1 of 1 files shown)


oskardotglobal (approved):
Thank you for your contribution. This is probably the best solution to the slow downloads.
Before we continue with this, please sign the CLA (so we can license your contribution under the AGPL). Also, I'd like to get opinions from some more contributors before we merge this.

LDprg (approved):
Nice, change removing privileged makes sense, since it should fix bugs with portainer. Also using 11 as default is probably the better way (People might not trust tiny11).

KernelGhost (approved):
LGTM as long as these changes have been tested 👍

PR #476, opened 2025-04-02
docs: Correct readme

kroese:
In the Readme for the Docker method it says:

```
After installing Windows, comment out the following lines in the `compose.yaml` file by prepending .....
```

and then it explains how to bind the `OEM` folder or a custom .iso, but I really don't understand why that was done.

Because after Windows installation has finished, setting the `OEM` folder has zero effect anymore, it only is used DURING installation. Same goes for the custom .iso.

There must be a reason why the creator of that document suggested to do it in a two-step process, where you don't add the `OEM` folder from the beginning, but I don't know the rationale behind it.

Maybe @KernelGhost can shed some light on why he took this approach?

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -56,6 +56,7 @@ docker compose --file ~/.config/winapps/compose.yaml down
 rm ~/.config/freerdp/server/127.0.0.1_3389.pem
 
 # Re-create the container with the updated configuration.
+# Add the -d flag at the end to run the container in the background.
 docker compose --file ~/.config/winapps/compose.yaml up
 ```
 
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -7,7 +7,7 @@ Although WinApps supports using `QEMU+KVM+libvirt` as a backend for running Wind
 > [!IMPORTANT]
 > WinApps does __NOT__ officially support versions of Windows prior to Windows 10. Despite this, it may be possible to achieve a successful installation with some additional experimentation. If you find a way to achieve this, please share your solution through a pull request for the benefit of other users.
 > Possible setup instructions for Windows 10:
-> - 'Professional', 'Enterprise' or 'Server' editions of Windows are required to run RDP applications. Windows 'Home' will __NOT__ suffice. It is recommended to download the ISO from a reputed source, as the built in downloader from dockur (default set to `tiny11`) will take longer than it would to download from a browser/torrent.
+> - 'Professional', 'Enterprise' or 'Server' editions of Windows are required to run RDP applications. Windows 'Home' will __NOT__ suffice. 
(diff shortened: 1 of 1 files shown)


oskardotglobal:
I see you've also included doc changes for #475, so we're gonna wait on that before merging this.

KernelGhost:
@kroese I believe the original instructions were written this way to avoid issues if the Windows installation ISO was deleted after setting up the container. That said, simplifying the instructions makes sense, especially if the only downside is the creation of a harmless OEM folder. Do we know how the container will behave if the ISO is still specified within `compose.yaml` but the file itself no longer exists at that path?

docs/docker.md
```diff
@@ -47,36 +44,18 @@ docker compose --file ./compose.yaml up
...
 
-After installing Windows, comment out the following lines in the `compose.yaml` file by prepending a '#':
-- `- ./oem:/oem`
-- `- /path/to/windows/install/media.iso:/custom.iso` (if relevant)
-
-Then, copy this modified `compose.yaml` file to `~/.config/winapps/compose.yaml`.
-
-```bash
-cp ./compose.yaml ~/.config/winapps/compose.yaml
-```
```
oskardotglobal, line 58: If we remove this, we still have to document where the user gets their `compose.yaml` from, especially the part that they should place it in `$XDG_CONFIG_HOME/winapps`. 

I would place something like this earlier into the file (above the Docker section, since this applies to both Docker and Podman):
````suggestion
First, grab a sample `compose.yaml` from our repository.

```bash
wget -O ~/.config/winapps/compose.yaml "https://raw.githubusercontent.com/winapps-org/winapps/refs/heads/main/compose.yaml"
```
````

Comment (oskardotglobal) [reply docs/docker.md:58]: You're right. The problem here is that the user is supposed to create the container *before* cloning the repository. I think the best solution for now is to leave it as is, and then adjust the installer accordingly

File: docs/docker.md [hunk 2]
```diff
@@ -47,36 +44,19 @@ docker compose --file ./compose.yaml up
 
 You can then access the Windows virtual machine via a VNC connection to complete the Windows setup by navigating to http://127.0.0.1:8006 in your web browser.
```
LDprg, line 45: When we use "container" instead of "virtual machine" we should be consistent.

docs/docker.md
```diff
@@ -110,35 +90,18 @@ podman-compose --file ./compose.yaml up
...
```
LDprg, line 91: See first comment.

LDprg (changes requested):
We should use 'container' and 'virtual machine'  consistent.

LDprg (approved):

oskardotglobal (approved):
LGTM.

PR #488, opened 2025-04-11
Share mounted media via /run/media

linull24:
Little fix for Dolphin user could directly open mounted files on portable devices like USB drive.
edit:
It is the default behavior fot udisks: [https://wiki.archlinux.org/title/Udisks#Mount_to_/media](archwiki).
But the FHS suggest [https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s11.html](url) using `/media`.
This version is released in 2015,10 years ago, so udisks doesn't follow it seems reasonable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -383,7 +383,13 @@ WAFLAVOR="docker"
 # - '140'
 # - '180'
 RDP_SCALE="100"
-
+# [MOUNTING REMOVEABLE PATHS FOR FILES]
+# NOTES:
+# udisks which you most probalbly have installed ,use /run/media by default.
+# This will improve the experience while using DEs.
+# However,FHS suggests using /media.Please check your system.
+# If you prefer to mount manually,you can change it to /mnt.
+REMOVABLE_MEDIA = "/run/media"
 # [ADDITIONAL FREERDP FLAGS & ARGUMENTS]
 # NOTES:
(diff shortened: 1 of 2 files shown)


oskardotglobal:
Is that standardized somehow or is that a dolphin exclusive?

linull24:
> Is that standardized somehow or is that a dolphin exclusive?

it is by default of `udisks` ,a XDG component.
However it is against FHS which suggest `/media`.
Anyway , the default case of udisks is more common,so I think it is OK.
Maybe we can add a varieble and update the document.

linull24:
Request for review.

README.md
```diff
@@ -383,7 +383,13 @@ WAFLAVOR="docker"
 # - '140'
 # - '180'
 RDP_SCALE="100"
-
+# [MOUNTING REMOVEABLE PATHS FOR FILES]
+# NOTES:
+# udisks which you most probalbly have installed ,use /run/media by default.
```
oskardotglobal, line 388: Can you fix the spaces before and after the commas and typos here?

oskardotglobal (changes requested):

README.md
linull24: down

oskardotglobal (approved):
Looks good.

PR #496, opened 2025-04-19
fix script runtime

thefiredragon:
Increase runtime to prevent issues when connecting to a self hosted windows vm.


oskardotglobal (changes requested):
This is not a good solution. Making the timeout configurable would be somewhat good, but the current solution makes no sense.

thefiredragon:
I can change this to make it to a proper variable. With defaults to 60 seconds.

If the remote machine needs more time we have here a bad error handling, so after script runtime a message should appear.

oskardotglobal:
This has been fixed in #524

PR #497, opened 2025-04-19
Document how to make rootless podman work

arwarw:
The documentation describes how to use podman like docker. But when using podman as a user to take advantage of its 'rootless' feature, a permission error accessing /dev/kvm will often be encountered.

This PR adds documentation about how to fix that problem.

See also https://github.com/containers/podman/blob/main/troubleshooting.md#20-passed-in-device-cant-be-accessed-in-rootless-container and https://github.com/containers/podman/issues/10166

v2: Fixed whitespace damage in previous cancelled PR https://github.com/winapps-org/winapps/pull/495 .


oskardotglobal (approved):
LGTM.

PR #517, opened 2025-05-24
Allow libvirt VM name change

sears-s:
Users may have a use case where they would like to change the libvirt VM name to one other than the default `RDPWindows`. For example, I may want to switch between using two different VMs for winapps, but two VMs cannot have the same name in libvirt. In the future, support could be added for multiple, simultaneous VMs. I've made the following changes:

- Removed the `readonly` attributed from `VM_NAME` for `winapps` and `winapps-setup`
- Updated documentation
- Made the error message for not finding the VM more descriptive

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,13 @@ RDP_DOMAIN=""
 # - 'libvirt': '' (BLANK)
 RDP_IP="127.0.0.1"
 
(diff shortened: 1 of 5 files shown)


oskardotglobal (approved):
Looks good.

PR #524, opened 2025-06-04
fix(setup.sh): increase timeout durations for RDP port check and process completion

### LINKED ISSUE #480: Time limits are too short
My winapps setup experienced time outs and issued incorrect error messages owing to too short timers in the code. Please adjust the time limits to 120 seconds.

setup.sh:
waCheckRDPAccess function:
\# Wait a maximum of 30 seconds for the background process to complete.
    while [ "$ELAPSED_TIME" -lt 30 ]; do

function waFindInstalled
\# Wait a maximum of 60 seconds for the batch script to finish running.
    while [ $ELAPSED_TIME -lt 60 ]; do

JoAllg:
For some users or systems, the time limits in `setup.sh` are too short, causing the script to fail. This pull requests increases the limits as reported working in #480.
fixes #480

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,6 +452,31 @@ AUTOPAUSE_TIME="300"
 # DEFAULT VALUE: '' (BLANK)
 # VALID VALUES: The command required to run FreeRDPv3 on your system (e.g., 'xfreerdp', 'xfreerdp3', etc.).
 FREERDP_COMMAND=""
+
+# [TIMEOUTS]
+# NOTES:
+# - These settings control various timeout durations within the WinApps setup. 
+# - Increasing the timeouts is only necessary if the corresponding errors occur. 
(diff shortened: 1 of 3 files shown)


JoAllg:
It does not seem appropriate to require organization access for the CLA.
edit: Fixed. It is an issue with disabled cookies or domain isolation (https://github.com/cla-assistant/cla-assistant/issues/566)
image

oskardotglobal (changes requested):
As discussed in #496, this is a bad solution since it highly depends on connection speed & hardware performance etc.

These values should be configurable.

JoAllg:
So do you suggest to add two configuration values to `~/.config/winapps/winapps.conf`?

oskardotglobal:
Yes, that is what I am suggesting.

JoAllg:
Please review the latest commit with configurable timeout settings and adapted README

PR #526, opened 2025-06-05
Remove more default RDP flags to simplify debugging

### LINKED ISSUE #512: Mouse not aligned with application
Hello have a strange issue when launch any application in Garuda linux the mouse isn't aligned with apllication when hover over a button or highlight something it's not in right place its usually a ways down from it. There is way to temp fix it by resetting my display scale settings small then large and app then functions normal I hope there's a fix I can apply for this cause really like Arch.

This is before when app is first launched

Image
Image

This is after when do display scale setting change and fixes the app

Image

### LINKED ISSUE #525: Changes to xfreerdp commandline
In the freerdp matrix room Akallabeth suggested that the flags "+clipboard -wallpaper +dynamic-resolution +span" be removed from the xfreerdp commandline for remote apps. The reasoning being that "+clipboard" is the default, "-wallpaper" does nothing for the remote app session, and both "+dynamic-resolution" and "+span" can lead to mouse events not being in the right spots. These sound like valid arguments to change the command line for remote app sessions.

Image

oskardotglobal:
According to the FreeRDP matrix (see #525), some of our default arguments are outdated.

I've gone ahead and removed these flags and they can always be re-added by the user using `RDP_FLAGS`.

One of those flags which was apparently causing issues is `+span`, so I've removed the `MULTIMON` config variable (which added span if it was false, and removing that it would add a flag with a conditional which seems redundant)

Closes #525

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,18 +403,9 @@ REMOVABLE_MEDIA="/run/media"
 # [ADDITIONAL FREERDP FLAGS & ARGUMENTS]
 # NOTES:
 # - You can try adding /network:lan to these flags in order to increase performance, however, some users have faced issues with this.
-# DEFAULT VALUE: '/cert:tofu /sound /microphone'
+# DEFAULT VALUE: '/cert:tofu /sound /microphone +home-drive'
 # VALID VALUES: See https://github.com/awakecoding/FreeRDP-Manuals/blob/master/User/FreeRDP-User-Manual.markdown
-RDP_FLAGS="/cert:tofu /sound /microphone"
-
-# [MULTIPLE MONITORS]
-# NOTES:
-# - If enabled, a FreeRDP bug *might* produce a black screen.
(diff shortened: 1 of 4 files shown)


theguywiththeproblem:
Shouldn't "+dynamic resolution" remain for the normal desktop RDP session? This flag is convenient for the desktop session and from my understanding should only be excluded when launching Remote Apps.

oskardotglobal:
@winapps-org/legacy can someone review?

theguywiththeproblem:
@oskardotglobal “+dynamic-resolution” should be retained for the desktop session (non-RAILS). This allows for the resizing of the window. The flag is not recommended for RAILS sessions only.

oskardotglobal:
No need to open an issue for that, I'll address it

LDprg (approved):
Seems perfect to me.

PR #528, opened 2025-06-05
doc: use openbsd-netcat on arch

### LINKED ISSUE #508: Suggestion for Arch Linux instructions in README
Maybe you could edit instruction for netcat to `openbsd-netcat` because that package on ArchLinux in my case CachyOS works. But for me Error is another: `Exiting with status '14'`. I replied for this [here](https://github.com/winapps-org/winapps/discussions/484)

oskardotglobal:
Title. Apparently, the `openbsd-netcat` package works better than `gnu-netcat` with our args on Arch. This is consistent with all our other packages except Fedora. Maybe we should install an openbsd version of netcat there too?

Closes #508

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,7 +311,7 @@ Install the required dependencies.
       ```
   - Arch Linux:
       ```bash
-      sudo pacman -Syu --needed -y curl dialog freerdp git iproute2 libnotify gnu-netcat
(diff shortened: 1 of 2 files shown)


oskardotglobal:
@winapps-org/legacy can someone review?

LDprg (approved):
Perfectly fine. Sorry for the late review.

Issue #275, opened 2024-09-26
Winapps Support for Mac


hykelvinlee42:
Hi there, if I understand the underlying mechanism correctly, Winapps utilizes virtual machines (e.g. `docker`) and `freerdp`, which both have support on MacOS. In theory, would the Winapps support running applications like Solidworks on MacOS?

oskardotglobal:
Well, technically it is possible, there is a POC on the `legacy-macos` branch

oskardotglobal:
However, especially Solidworks is something that would run really poorly in a VM

hykelvinlee42:
Thanks, I'm not planning to use Solidworks in particular. It's just the first Windows only application that came to my mind.

KernelGhost:
It is *technically* possible to add macOS support through `utmctl` in a way similar to how `libvirt` is currently utilised. I can explore this on my M1 MacBook Pro, but I can’t commit to providing a solution as stability may be an issue. I’ll try to revisit this when I have more time in the coming months.

<img width="3360" height="2100" alt="Image" src="https://github.com/user-attachments/assets/008cbffb-08b9-485b-bdce-fa6ae718349c" />

iMonZ:
> mit to providing a solution as stability may be an issue. I’ll try to revisit this when I have more time in the coming months.

Shouldn't docker just work fine for now?
I think the Shell or UX integration would be more important than the actual virtualization of Windows.
The arm version of Windows would work fine as well for now.

@KernelGhost How did you get your demo running on the screenshot?

KernelGhost:
@iMonZ I ran the Windows VM inside UTM with TCP and UDP traffic forwarded to the host on port 3389. FreeRDP was installed via [Homebrew](https://formulae.brew.sh/formula/freerdp#default). While it generally worked, dragging windows was unreliable (only window snapping seemed to function properly). The setup also felt a bit fragile as FreeRDP would occasionally crash.

iMonZ:
> [@iMonZ](https://github.com/iMonZ) I ran the Windows VM inside UTM with TCP and UDP traffic forwarded to the host on port 3389. FreeRDP was installed via [Homebrew](https://formulae.brew.sh/formula/freerdp#default). While it generally worked, dragging windows was unreliable (only window snapping seemed to function properly). The setup also felt a bit fragile as FreeRDP would occasionally crash.

I see that you use XQuarz. Is there maybe another, better alternative?

KernelGhost:
[...]

Issue #43, opened 2023-09-24
Add support for switching input method


FengZhongShaoNian:
I'm using the Chinese version of windows 10 , which uses the system's built-in pinyin input method. I use `./bin/winapps manual Notepad.exe` to open notepad. I want to input some Chinese in Notepad, but I can't switch to the Chinese input method.

Technical details:

Host OS: Manjaro Linux with KDE plasma 5.27.8
Winapps: Latest git.
FreeRDP packages: freerdp 2:2.11.1-1
Arch: amd64.
Guest OS: Windows 10 PRO 22H2（Chinese Version）.

Input method on Linux: fcitx5-rime
Please let me know if you need any more information.
If there is some configuration I can do on my end to get this working properly, also please let me know.

Thank you and congratulations for this awesome project!

oskardotglobal:
Have you tried connecting to the VM normally via RDP and changing the Keyboard layout in windows?
Otherwise, maybe you'll find something useful here: https://github.com/FreeRDP/FreeRDP/issues/8980

FengZhongShaoNian:
![Screenshot_20230924_184040](https://github.com/winapps-org/winapps/assets/16934087/62c499f5-5e83-415c-abf8-d9031611e8aa)

As you can see above, I have two keyboard layouts on my windows, and I tried to remove the English one. Now after opening notepad by  `./bin/winapps manual Notepad.exe `, I can see the candidate word panel of the Chinese Pinyin input method, but it flashed away during the input process (as shown below), so I still can't enter Chinese normally.

![Peek 2023-09-24 18-58](https://github.com/winapps-org/winapps/assets/16934087/698da411-9132-4ab9-b4a4-30d9b452d47d)

LDprg:
Is your System layout also Chinese? Maybe thats a problem.

ctrngk:
Do not use any default input method offered by windows.  Use 3rd party software like baidu, which includes a feature to trigger the input method by customizing your own shortcut.

jrevillard:
Hello,

I have the exact same problem with the French Keyboard. Since the migration to Docker Windows image, I don't know exactly why but it seems that the US keyboard cannot be removed.... When I start any application, this is always the one selected. I tried so many things to remove this keyboard layout or to ask FreeRDP to use the French Keyboard that I don't know what to do anymore.....

Best,
Jerome

LDprg:
[...]

Issue #34, opened 2023-08-08
Add support for Tray icons


upintheairsheep:
Hello, can you add support for mirroring the Windows taskbar tray icons that indicate stuff to the Linux tray, for parity with Parallels Desktop?

oskardotglobal:
Hello, this is not a straight-forward task and is probably not possible without custom software on both client and server. This might be added in the rewrite, but it's not a priority and it's entirely not possible with the legacy shellscript

PR #544, opened 2025-06-10
Remove Launcher submodule

oskardotglobal:
Title.

I have no idea why we added it in the first place, it isn't very helpful.


ghost:
At the current moment it is in the code of main branch and is redirecting to other branch. I think it would be good to get that launcher into main branch at first and then decide.

ghost:
When I looked at it at first I thought that It is a programm for windows to do some things like file sharing and etc, but then I seen linux and really had to think why it is needed.

ghost:
I think launcher can be used for #34 .
Display tray icons that are in windows to Linux tray.

ghost:
#43 I think that tray launcher can be good thing to add input method switch in windows.
Cuz when I am in Linux and try to switch input method it only does switch it in Linux even if I am in MS Word and it is focused, it still only change in Linux. SO having option in tray to switch windows input method would be good.

ghost:
In #547 I notet that launcer could be used to start and close container without terminal.

oskardotglobal:
This is about having the submodule. The launcher can continue living on its own repository.

ghost:
Oh then this is great.

LDprg:
Maybe we could add some docs about its existence. That would be better than the submodule.

oskardotglobal:
Yeah, that would be good
@LDprg can you review?

oskardotglobal:
Reminder to self: Remove the `--recurse-submodules` from setup.sh

KernelGhost:
I agree with the above discussion points. Removing the submodule and documenting the launcher's existence in the README is a better approach.

Some additional thoughts:
1. The launcher is built on [`yad`](https://github.com/v1cont/yad), which [lacks native Wayland support](https://github.com/v1cont/yad/issues/159). This could become problematic in the future, as XWayland (a bridging technology intended to ease the transition from X11 to Wayland) is ultimately a stopgap and may not be maintained indefinitely.
2. I originally wrote the launcher whilst using XFCE, and it was only ever optimised for that environment. The launcher has become increasingly difficult to use on some desktop environments like GNOME, which has moved away from supporting taskbar notifications due to UI/UX decisions. Although GNOME extensions can partly restore functionality, they introduce the risk of inconsistent or unpredictable behaviour. I'm unsure how well the launcher performs under KDE Plasma.
3. `yad` was initially chosen as a quick and simple way to build a GUI with bash, but in hindsight, it may not have been the most future-proof choice. For example, [issue #275](https://github.com/v1cont/yad/issues/275) continues to impact the launcher's ability to automatically detect the status of the Windows VM, requiring inclusion of a manual refresh button within the launcher.

At some point, it may be worth considering rebuilding the launcher using a more modern toolkit. That said, this is probably a broader conversation better suited to a dedicated thread.

I'm happy for the proposed changes to proceed.

KernelGhost:
I have updated the README in 7188ed4 to include a section on the launcher. Several mentions to the launcher remain in the Nix-related sections, but I don't think I'm the best person to modify these.

oskardotglobal:
> This could become problematic in the future, as XWayland (a bridging technology intended to ease the transition from X11 to Wayland) is ultimately a stopgap and may not be maintained indefinitely.

Yes, especially since Gnome & KDE are phasing out X in favour of wayland 

> At some point, it may be worth considering rebuilding the launcher using a more modern toolkit. That said, this is probably a broader conversation better suited to a dedicated thread.

We planned to have a GUI with the rewrite anyways, so that could go there
Also, since there'd be a common lib you wouldn't even have to call CLI tools by hand anymore

ghost:
I am making GUI Install Assistant using c# and AvaloniaUI. So maybe I could look at launcher to rewrite it in more GUIsh way and integrate it in my installer. This is not going to conflict with rewrite as I see that it is progressing slow and I do not touch the shell scripts. And I am only making a Installer of sort so users do not need to read tons of text in readme.

PR #546, opened 2025-06-10
Remove fonts

oskardotglobal:
The font in the repository is only used in the icon and can be safely removed. It's not installed by any code.


ghost:
https://github.com/winapps-org/winapps/blob/main/icons/winapps.svg
This is as well nowhere to be seen in the code. But may be usefull if you do want to develop own app.

I propose to move this icon and fonts to this branch instead or removing
https://github.com/winapps-org/winapps/tree/rewrite

Maybe it will get usefull.

oskardotglobal:
We'd always have it in the Git history anyways.

ghost:
It is just that it will be waiting in rewrite branch readily, while after deletion we need to actually search for stuff.

oskardotglobal:
So it'll clutter the rewrite branch, that doesn't help
We could make an assets branch though where we keep stuff like this

ghost:
> So it'll clutter the rewrite branch, that doesn't help We could make an assets branch though where we keep stuff like this

Yes, that would be the best option. Always ready to merge any icon or font whenever needed.

LDprg:
Well, doesn't this PR itself do enough for preservation? We could even tag this PR with a special archive tag or something similarl. This way we prevent any garbage in the git tree.

LDprg (approved):

PR #565, opened 2025-06-20
Wait a bit before running command if Windows is not booted

eylenburg:
In my experience with WinApps, if you start an app but the Windows VM has not booted, it just says "Booting Windows..." but then nothing happens and you have to start the app a second time. This commit would add a short waiting period before running the freerdp command so that hopefully the Windows VM is ready by then.

### DIFF
--- a/bin/winapps
+++ b/bin/winapps
@@ -55,6 +55,7 @@ MULTI_FLAG=""
 
 # OTHER
 FREERDP_PID=-1
+NEEDED_BOOT=false
 
 ### TRAPS ###
 # Catch SIGINT (CTRL+C) to call 'waCleanUp'.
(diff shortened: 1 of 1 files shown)


bin/winapps
```diff
@@ -412,6 +413,8 @@ function waCheckContainerRunning() {
     local TIME_ELAPSED=0
     local TIME_LIMIT=60
     local TIME_INTERVAL=5
+    local MAX_WAIT_TIME=120
```
oskardotglobal, line 416: This timeout should probably be configurable, like with #524

bin/winapps
```diff
@@ -698,7 +742,12 @@ else
 fi
 
 waCheckPortOpen
-waRunCommand "$@"
+if [ "$NEEDED_BOOT" = "true" ]; then
+    sleep 10
```
oskardotglobal, line 746: This sleep statement along with the `NEEDED_BOOT` seems somewhat arbitrary, since you check the container state extensively anyways?

bin/winapps
eylenburg: I added it as a global variable and in the .conf file

PR #574, opened 2025-06-23
Add a function to sync time in Windows after Linux host is suspended (sleeping)

eylenburg:
I noticed this issue that the time in the Windows VM is "frozen" while the Linux host is sleeping (suspended). So will end up with a wrong time in the Windows VM. This might be fine for most applications, but it also affects file modification dates, which, if you edit a file both in a Linux app and with a WinApp, can lead to conflicts regarding which version of a file is newer.

This PR is an attempt to fix it. It's not the most elegant of solutions so I would understand if you don't want to merge it, but I just wanted to share it. Perhaps there is a much better and elegant solution - after all Windows has an automatic time sync in the Settings app, but it just doesn't seem to work for me.

What it does is after Linux wakes up from sleep, the next time WinApps is used it creates an empty file (the sleep marker) in the APPDATA folder and at the same time there is a script running in the background in Windows that checks every 5 minutes (arbitrary time) whether the sleep marker exists (which should be accessible via RDP). If yes, it will resync the time and delete the sleep marker.

This would require a script running in the Windows VM, which is added in the oem folder, so it would only work with newly set up VMs.


LDprg:
Thx @eylenburg for your contribution.
Please fix precommit issues (see ci/cd).

theguywiththeproblem:
Not meaning to be critical but  the hacky solution I use for my system is appending “virsh domtime ‘VM_Name’ —now” after every instance of virsh resume in the core winapps script and the launcher. I’m not sure if this would have issues with podman or docker as I use a libvirt vm.

AkechiShiro:
Hi @oskardotglobal @LDprg, do you think this PR is still relevant ? Should we try to move forward with it, I see it has a conflict right now, but I believe the precommit issue is fixed.

LDprg:
As far as I know this is still relevant. The merge conflict is probably just a new  function from another pr.

AkechiShiro:
@eylenburg could you please fix the conflict then we could probably can merge this PR, if there isn't anything left to address so this issue is fixed.

eylenburg:
I've synced the fork and fixed an error I spotted

AkechiShiro (changes requested):
Could you please sign the DCO @eylenburg ? I believe you need to add signed off into the commit see instructions here : https://github.com/winapps-org/winapps/pull/574/checks?check_run_id=49318164595

eylenburg:
Ok I wasted an hour trying to figure it out, giving up now. Git just tells me `Successfully rebased and updated detached HEAD.` and `Everything up-to-date` but I can't see any sign-off on github.com.

Just use my PR as you want, I hereby license it as AGPL or MIT or WTFPL or whatever you like :)

oskardotglobal:
Here, I've gone ahead and signed off the commits on your behalf. FYI, the way you do this is normally is by adding `--signoff` (or `-s`) to your `git commit` calls. Retroactively, you can do it by `git rebase HEAD~<number of commits> --signoff`, but that doesn't work if there's been merge commits, so what I did is roll my local branch back to your latest commit, then rebase to sign off, and then merge to main.

PR #597, opened 2025-07-07
Add README notice for non-English inputs options

denisstrizhkin:
I am using both Russian and English in MS Word. So it took me a while to figure out how to make the VM copy client's inputs. I think this can be helpful for other users to know.


oskardotglobal (approved):
LGTM

PR #620, opened 2025-07-31
fix(nix): include icons dir for setup script

wovw:
Added `icons` dir to nix derivation so `winapps-setup` (installed through the nix flake) succeeds


oskardotglobal (approved):
Thanks, I missed that

PR #641, opened 2025-08-19
Update README.md with note for Debian 12

9Morello:
This is a minor PR that updates the `README` to make it clear the Debian instructions for installing `freerdp3-x11` are for Debian 12 (Bookworm). The current stable release (Trixie) has this package on its default repositories, so using backports isn't necessary anymore

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,7 +303,7 @@ Install the required dependencies.
       ```
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


oskardotglobal (approved):
TIL Trixie is stable now. Thank you for your contribution.

PR #663, opened 2025-08-25
Improve libvirt documentation and install script

KernelGhost:
Summary of Changes
- Automated Remote Desktop enablement (via 'install.bat' and 'RDPApps.reg')
- 'install.bat' now executes end-to-end without requiring user input (potentially important for unattended container setup)
- 'libvirt' documentation updates
    - Removed instruction to manually enable Remote Desktop (now automated, as above)
    - Optional steps are now collapsed/hidden by default for clarity
    - VirtIO installation instructions simplified
    - Removed redundant "Rename this PC" step
    - Updated screenshots

Requesting review with particular attention to any implications for podman/docker backends, as I believe these still run 'install.bat' when created for the first time as per [this](https://github.com/dockur/windows?tab=readme-ov-file#how-do-i-run-a-script-after-installation).

### DIFF
--- a/oem/install.bat
+++ b/oem/install.bat
@@ -77,11 +77,3 @@ if %ERRORLEVEL% neq 0 (
         echo [ERROR] Failed to create scheduled task "%taskname%".
     )
 )
-
-echo.
-echo Exiting in 10 seconds...
-echo Press Ctrl+C to pause.
-for /l %%i in (10,-1,1) do (
-    <nul set /p="%%i... "
-    timeout /t 1 >nul
-)


oem/install.bat
```diff
@@ -1,33 +1,58 @@
...
 
-REM Confirm the user wants to proceed with setup
+:: Confirm the user wants to proceed with setup
 echo ============================================
 echo             WinApps Setup Wizard
 echo ============================================
 echo.
-echo Press any key to continue or close this window to cancel...
-pause >nul
+echo Waiting 5 seconds before starting...
```
oskardotglobal, line 16: Since this is unattended like half of the time, I think these waits are unnecessary

oskardotglobal (changes requested):

oem/install.bat
KernelGhost: Agreed, however I have retained the 10 second post-execution delay as it allows time for users to review the script output for warnings and errors, and to make a copy of this for debugging purposes.

PR #675, opened 2025-08-27
feat: add Adobe CC 2024 integrations

queenkjuul:
The new entries (AME, Premiere) should work for 2022 as well (and maybe also 2018) just by changing the years, but I don't have the old versions around to test anymore. 

I have not verified the MIME type for Premiere Pro project files, I am not sure how to do that. I just manually set the file handler in my DE and that works fine.


oskardotglobal (approved):

oskardotglobal:
Thank you for your contribution.

PR #680, opened 2025-08-29
fix(nix): set write permissions when copying from nix store

Dreamail:
fix #664


oskardotglobal (approved):
Thank you for your contribution.

Issue #638, opened 2025-08-16
similarly named kvm instances


queler:
https://github.com/winapps-org/winapps/blob/c5bdc1eee433d646ade951ce59ae15a0b052253f/setup.sh#L844C3-L854C53

Found a fun little bug.  i was troubleshooting, and trying to upgrade my KVM, so i wound up with a bunch of kvm's that were off named things like RDPWindows.backup and RDPWindowsorignal.  The problem comes at 
`   
# Obtain VM Status
    VM_PAUSED=0
    virsh list --state-paused | grep -wq "$VM_NAME" || VM_PAUSED="$?"
    VM_RUNNING=0
    virsh list --state-running | grep -wq "$VM_NAME" || VM_RUNNING="$?"
    VM_SHUTOFF=0
    virsh list --state-shutoff | grep -wq "$VM_NAME" || VM_SHUTOFF="$?"

    if [[ $VM_SHUTOFF == "0" ]]; then
        # Complete the previous line.
        echo -e "${FAIL_TEXT}Failed!${CLEAR_TEXT}\n"`

it then detects the running AND shutoff machines.  Since the first test is for power off, the script thinks the VM is off

KernelGhost:
Just to clarify, is the bug due to detection of the '.' in 'RDPWindows.backup' as a word boundary, thus returning the shutoff status for 'RDPWindows.backup' instead of 'RDPWindows'?

queler:
That's right. I started to fix it with a combination of awk, virsh dominfo, and/or surrounding the regexes with \s instead of word-regex. Spaces don't appear to be legal in kvm. Unfortunately, I really needed to be done with my project and my first round of attempts didn't work, so I gave up and just renamed them.
The problem does occur in both the install and main script

PR #707, opened 2025-09-04
Fix: #697 and #638

### LINKED ISSUE #697: Script Erroneously Claiming that the VM is not Powered On.
I don't really know how else to describe this issue other than it's erroneously claiming that the VM is powered off when, clearly, it's running. It's also not creating the log file so I can't provide any more info than this.

Image****

KernelGhost:
Fix #697 and #638 using a more robust string comparison method.


oskardotglobal:
Looks good.

PR #722, opened 2025-09-08
Add Capture One and update adobe apps

Arakmar:
Creative Cloud is now installing the 2025 version of some Adobe apps

Their naming scheme is really a mess...
Maybe some old Adobe app versions can be removed ?


LDprg:
@oskardotglobal I am not sure about just adding new Adobe stuff. They tend to change they naming from time to time. Replacing don't seem like an option since old Adobe versions might still be used out there (Adobe doesn't deploy new versions to every user at the same time). However I am not sure, whether there is a better way to do this than just keep two Adobe versions around.

Arakmar:
@LDprg I agree, it will be complex to maintain if they decide to change the path every year...
Just an idea, maybe it could be easier to have multiple possible path in the same info file for a specific application ? (with a bash array or something similar)

LDprg:
Bash Arrays seem to be a good solution for this. This way we could add alternative paths, which is not a problem, since you won't install multiple versions.

@oskardotglobal what do you think?

oskardotglobal:
Arrays are a good idea, the only issue with that would be if a user has two versions

Issue #444, opened 2025-03-01
Podman way of running windows , windows adverstises its IP as 20.20.20.21 , & RDP doesnt work


gitnohubz:
Fedora 41 using the podman way , i can easiky accees windows via vnc in the browser , but no success at rdp session , the advertised IP is weird & xfreerdp commans simply fails miserably ):

KernelGhost:
@gitnohubz Did you preface your test command with `podman unshare --rootless-netns`? For example, `podman unshare --rootless-netns xfreerdp /u:"YourWindowsUser" /p:"YourWindowsPassword" /v:"YourWindowsVMIP"`? This is necessary to join the network namespace associated with rootless Podman containers.

@oskardotglobal I note that this specific instruction is missing from our [README](https://github.com/winapps-org/winapps?tab=readme-ov-file#step-4-test-freerdp) in the section where we instruct users to test their configuration, despite this being correctly implemented in both `setup.sh` and `bin/winapps`. It might be worth updating the README to specify using this particular command prefix for podman users.

Issue #251, opened 2024-09-10
[Upstream issue] Windows 11 + Podman : RDP seems to not work   


AkechiShiro:
We tried to use `Podman` but RDP connection seemed to fail, this is most likely related to this upstream issue :  https://github.com/dockur/windows/issues/679

DaRandomCube:
so, you can access it via vnc, but not rdp? (or atleast, rdp doesn't work anymore)

AkechiShiro:
Yes via vnc it works

DaRandomCube:
> Yes via vnc it works, docker works with VNC and RDP @DaRandomCube

we have the same thing then
anyway any idea on how to fully delete the win11 podman thing? as when i use `podman-compose` rm and try to `podman-compose --file`, it doesn't download the windows 10 iso, aka there are some leftovers

AkechiShiro:
Maybe this : https://docs.podman.io/en/latest/markdown/podman-system-prune.1.html

I haven't tested but I need to cleanup as well

AkechiShiro:
I think I'll try to capture with wireshark at a later time a successful authentification with Docker and the issue with Podman and then try and compare both to see if I can pinpoint what is the issue.

DaRandomCube:
> I think I'll try to capture with wireshark at a later time a successful authentification with Docker and the issue with Podman and then try and compare both to see if I can pinpoint what is the issue.

any updates?

AkechiShiro:
Nope I didn't find time to do it, assume there is no news unless I put an update here publicly

AkechiShiro:
To quote my answer upstream, a workaround is the following and then `winapps` can be used inside the shell spawned with the `podman` command.

> That seems to work @Moon1moon but only in the shell spawned from
> 
> ```
> podman unshare --rootless-netns
> ```
> 
> Then inside that shell, you can connect using `xfreerdp` and it does work with podman.
> 
> I had to specify (under NixOS with a shell managed by Home-Manager, else it was failing to enter the new shell)
> 
> ```
> SHELL=/bin/sh podman unshare --rootless-netns
> ```
> 
> with the following error :
> 
> ```
> Error: fork/exec /run/current-system/sw/bin/zsh: no such file or directory
> ```

DaRandomCube:
> To quote my answer upstream, a workaround is the following and then `winapps` can be used inside the shell spawned with the `podman` command.
> 
> You can try @DaRandomCube, should work for you
> 
[...]

PR #737, opened 2025-09-11
Update README with correct FreeRDP test command for Podman

KernelGhost:
I came across this issue while reading through #444. It may or may not address the issue, but since I don’t use Podman I’d appreciate it if someone could confirm whether the correction is accurate.


KernelGhost:
Further Context: https://github.com/winapps-org/winapps/issues/444#issuecomment-3282935048
Related Issue: [#251](https://github.com/winapps-org/winapps/issues/251)

PR #748, opened 2025-09-17
Update docker.md

ANDev-eg:
added `[]` to podman setup.

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -82,7 +82,7 @@ Please follow the [`docker` instructions](#setup-docker-container).
 > #### Rootless `podman` containers
 > If you are invoking podman as a user, your container will be "rootless". This can be desirable as a security feature. However, you may encounter an error about missing permissions to /dev/kvm as a consequence.
 >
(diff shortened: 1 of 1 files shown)


oskardotglobal (approved):

eylenburg:
I don't understand the updated readme. Isn't

```
group_add:
  - keep-groups
  ```
  what you need to have in your `compose.yaml` for rootless podman to work, i.e. uncomment the two lines like it said before?
  
The readme change is a bit confusing for me, if you take it literally ("uncomment the `group_add:` section at the end, and add `[]`.") it would now read, depending on how you interpret it:
  
  ```
group_add:
  - keep-groups []
  ```

or

  ```
group_add: []
  - keep-groups
  ```

or maybe

  ```
group_add: []
#  - keep-groups
  ```

or even

  ```
group_add:
  - keep-groups 
[]
  ```

**Comment by ANDev-eg** [comment]
Sorry for the late reply, I removed the setup because I'm testing the installation in another method.

This is the right one:

```shell
group_add:
- keep-groups []
```

oskardotglobal:
I seem to have read this wrongly during review, the change doesn't seem to make any sense

oskardotglobal:
I've done some reading and apparently, the compose spec does not support the keep-groups property, but since podman uses `crun` we can use
```yaml
annotations:
  run.oci.keep_original_groups: 1
```
Can someone confirm this?

ANDev-eg:
It worked for me on Ubuntu 24.04 when I was installing using podman. made sense to my PC.

Issue #396, opened 2025-01-28
Fail to launch apps when rdp certificate is renewed


freechelmi:
How to reproduce : 

- Install and winapps during 3 months
- at some point the certificate will ne renewed by Windows 
- after that , when launching a winapps app , you won't get any feedback because freerdp wants you to confirm you accept the new cert .

We need some way either for freerdp to always accept new cert or detect it and import it on the linux side

```
flatpak run --command=xfreerdp com.freerdp.FreeRDP /d: /u:Docker /p:Docker /scale:100 +auto-reconnect +clipboard +home-drive /audio-mode:1 -wallpaper +dynamic-resolution +span '/wm-class:Microsoft Word' '/app:program:C:\Program Files\Microsoft Office\root\Office16\WINWORD.EXE,icon:/home/michel/.local/share/winapps/apps/word-o365/icon.svg,name:Microsoft Word' /v:127.0.0.1
[13:50:44:787] [2:00000003] [WARN][com.freerdp.crypto] - [verify_cb]: Certificate verification failure 'self-signed certificate (18)' at stack position 0
[13:50:44:787] [2:00000003] [WARN][com.freerdp.crypto] - [verify_cb]: CN = DOCKERW-J03S567
[13:50:44:788] [2:00000003] [ERROR][com.freerdp.crypto] - [tls_print_certificate_error]: New host key for 127.0.0.1:3389
[13:50:44:788] [2:00000003] [ERROR][com.freerdp.crypto] - [tls_print_certificate_error]: @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
[13:50:44:788] [2:00000003] [ERROR][com.freerdp.crypto] - [tls_print_certificate_error]: @    WARNING: NEW HOST IDENTIFICATION!     @
[13:50:44:788] [2:00000003] [ERROR][com.freerdp.crypto] - [tls_print_certificate_error]: @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
[13:50:44:788] [2:00000003] [ERROR][com.freerdp.crypto] - [tls_print_certificate_error]: The fingerprint for the host key sent by the remote host is 15:71:75:c8:42:18:21:bd:55:f5:a5:63:a3:e4:7d:83:0e:bd:61:85:5e:46:36:f6:b7:85:1b:e7:11:b2:c2:95
[...]

PR #753, opened 2025-09-19
Automatically update rdp cert

heros-os:
Automatically Update RDP certificate before launching FreeRDP

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Ensure Windows-friendly line endings and encoding for batch and registry files
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.reg text eol=crlf working-tree-encoding=UTF-16LE-BOM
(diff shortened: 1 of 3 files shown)


KernelGhost:
@heros-os Thanks for this. Am I right in thinking this behaves the same as simply using `/cert:ignore` instead of `/cert:tofu` in `~/.config/winapps/winapps.conf`? I also noticed you hard-coded `CERT_PATH="$HOME/.var/app/com.freerdp.FreeRDP/config/freerdp/server/127.0.0.1_3389.pem"`, which may not apply to all FreeRDP installs.

Could you explain why this step is needed? I recall reading that RDP certificates can rotate automatically [here](https://github.com/winapps-org/winapps/issues/396), but I’m not sure how often this happens. Wouldn’t it be simpler either to rely on `/cert:ignore`, or (if Windows does regenerate RDP certificates) to add a registry tweak in `RDPApps.reg` to prevent this? Otherwise, overwriting the `.pem` with whatever certificate Windows presents seems equivalent to ignoring it? Keen to hear your thoughts.

Issue #730, opened 2025-09-10
can't run install.bat with run as administrator


2112992430:
### What happened?

i ran install.bat with run as administrator,but it keep closing it and opening it again and again,also it showed about e is not a command and check is not a command 

### Your FreeRDP version and where you got it from

3.1

### Your Linux distribution and version

Arch Linux

### Your `winapps.conf`

```shell
no winapps.conf
```

### Logs

```shell

```

### Terms

- [x] I am running the latest version.
- [x] To the best of my knowledge, this is a bug and not a setup nor a FreeRDP problem.
- [x] I have checked for duplicate issues.
- [x] I agree to follow this project's Code of Conduct.

KernelGhost:
@2112992430 Is there any way you can screenshot or copy the error messages you are receiving?

dragon-Elec:
It should be easily dable, which type of windows are you using is it a modified build ?, can you run the install.bat in Power shell?

I suggest running the cmd directly with admin to avoid this.

2112992430:
ive already run it with admin

2112992430:
> [@2112992430](https://github.com/2112992430) Is there any way you can screenshot or copy the error messages you are receiving?你能找到任何方法截屏或复制你收到的错误信息吗？

no there isnt,just like i said

dragon-Elec:
> ive already run it with admin

Open start memu, type cmd right click, click run as admin on the cmd icon menu, drag the script make sure 3 of the files in save folder or dir, hit enter see if there is any output in cmd terminal if there is no output hit enter again and let the cmd do it's job.

2112992430:
> > ive already run it with admin我已经用管理员权限运行了它
> 
> Open start memu, type cmd right click, click run as admin on the cmd icon menu, drag the script make sure 3 of the files in save folder or dir, hit enter see if there is any output in cmd terminal if there is no output hit enter again and let the cmd do it's job.打开开始菜单，输入 cmd 右键，在 cmd 图标菜单中点击以管理员身份运行，拖动脚本确保保存文件夹或目录中有 3 个文件，按回车查看 cmd 终端是否有输出，如果没有输出再次按回车，让 cmd 完成它的任务。

nothing happened,it still like that i said

dragon-Elec:
Weird this shouldn't happen, can tell which windows u are using? Is it a modified windows?

2112992430:
[...]

PR #754, opened 2025-09-19
Add .gitattributes + Fix encodings

KernelGhost:
This PR addresses [issue #730](https://github.com/winapps-org/winapps/issues/730), which (I assume) was caused by incorrect handling of Windows file encodings and line endings.

Changes:
- Enforce CRLF line endings for all Windows-related files (`.bat`, `.cmd` and `.reg`) via `.gitattributes`
- Normalise `.reg` files to UTF-16LE with BOM, which is the expected format
- Remove unicode characters from `install.bat`, ensuring it is pure ASCII

Tested on a Windows 11 VM. Batch scripts execute as expected and registry imports succeed without errors.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# Ensure Windows-friendly line endings and encoding for batch and registry files
+*.bat text eol=crlf
+*.cmd text eol=crlf
+*.reg text eol=crlf working-tree-encoding=UTF-16LE-BOM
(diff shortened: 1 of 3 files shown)


oskardotglobal (approved):
Looks good. Do we still have to manually reencode the files as crlf?

KernelGhost:
@oskardotglobal I don't think so - I cloned my branch and confirmed that the encodings and line endings were correct.

oskardotglobal:
Perfect.

PR #757, opened 2025-09-22
added ArcGIS Pro

Mammutor:
added ArcGIS Pro as a supported application and tested it locally. I also ran pre-commit and it passed everything

The icon originates from: https://en.wikipedia.org/wiki/File:Arcgis_Pro_logo.svg
I dont know if its usage is fair use, you can decide

### DIFF
--- a/apps/arcgis-pro/info
+++ b/apps/arcgis-pro/info
@@ -11,4 +11,4 @@ WIN_EXECUTABLE="C:\Program Files\ArcGIS\Pro\bin\ArcGISPro.exe"
 CATEGORIES="WinApps;GIS"
 
 # GNOME mimetypes
-MIME_TYPES=""
+MIME_TYPES=" "


oskardotglobal:
Please signoff the commit

Mammutor:
i am unable to to this, sorry

LDprg:
@Mammutor Do you need help for signing off the commit? Since it seems like you tried, based on the commit history.

[Github](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits ) has good docs about this. Also most gui git clients like Gitkraken provide simpler ways to do this (see the according docs for it).

Another way to sign commits is to create them within the github online editor (which only works good for single file edits).

If you need further help just answer this comment. If you have question about other topics regarding this Pr also feel free to ask them. It would be a shame if your Pr won't get merged, because of this.

PR #805, opened 2025-11-02
Renamed compose.yaml to docker-compose.yml everywhere

dhjensen:
Applied `dclint . --fix` -> https://github.com/zavoloklom/docker-compose-linter


oskardotglobal:
No. `compose.yaml` is a well-defined name, especially when using Podman.

PR #808, opened 2025-11-02
Add Affinity (3.0) by Canva app

TheChilliPL:
Adds Affinity 3.0 by Canva. Seems to work fine, has an SVG icon taken from Wikipedia

Hopefully fixable in the future:
- There seem to be a lot of transparent Windows opening when Affinity is started. They all disappear within a few seconds.
- Sometimes I need to switch the window several times for it to start recognizing mouse clicks. Happens with file explorer too, probably not app-specific.

### DIFF
--- a/apps/affinity/info
+++ b/apps/affinity/info
@@ -1,8 +1,3 @@
-# Copyright (c) 2024 Fmstrat
-# All rights reserved.
-#
-# SPDX-License-Identifier: Proprietary
-
 # GNOME shortcut name
 NAME="Affinity"
 


apps/affinity/info
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2024 Fmstrat
+# All rights reserved.
+#
```
oskardotglobal: See contributing guide, remove this

oskardotglobal (approved):
Looks good. Thank you for your contribution.

paidlowered:
Added README entry in #813

PR #813, opened 2025-11-04
Add Affinity by Canva to README.md

paidlowered:
Restructuring the table is a pain in the ass, I don't know how but we should switch to something different


paidlowered:
#808

oskardotglobal:
I don‘t know if Github readmes support CSS grid, if so that would be an option

Otherwise we could just generate the table with a script

oskardotglobal (approved):

PR #820, opened 2025-11-14
Added ChemDraw Prime

TheHeadlessTurtle:
I used the latest version of ChemDraw for this, which is ChemDraw Prime 25.0.2. I could not test other parts of the ChemDraw suite as my license would not let me activate those for some reason, even though I've previously activated other parts of the suite without problems (this almost certainly has nothing to do with WinApps or the Windows VM).

I apologize if I've done anything wrong here, this is the first time I'm using GitHub for anything other than downloading software.

### DIFF
--- a/apps/ChemDraw/info
+++ b/apps/ChemDraw/info
@@ -11,4 +11,4 @@ WIN_EXECUTABLE="C:\Program Files\RevvitySignalsSoftware\ChemDrawApplications_x64
 CATEGORIES="WinApps;Science"
 
 # GNOME mimetypes; file types: .cdxml .cdx .chm
-MIME_TYPES="chemical/x-cdx;application/vnd.chemdraw+xml;application/vnd.ms-htmlhelp;"
+MIME_TYPES="chemical/x-cdx;application/vnd.chemdraw+xml;chemical/x-chemdraw;"


apps/ChemDraw/info
```diff
@@ -0,0 +1,14 @@
...
+FULL_NAME="ChemDraw Prime"
+
+# The executable inside windows
+WIN_EXECUTABLE="C:\Program Files\RevvitySignalsSoftware\ChemDrawApplications_x64\ChemDraw.exe"
+
+# GNOME categories
+CATEGORIES="WinApps;Science"
+
+# GNOME mimetypes; file types: .cdxml .cdx .chm
+MIME_TYPES="chemical/x-cdx;application/vnd.chemdraw+xml;application/vnd.ms-htmlhelp;"
```
oskardotglobal: Is the `help` intended at the end here?

TheHeadlessTurtle: It was intended, but while checking if I had made a mistake there I actually found a total of three different mimetypes for .chm files... Those are `application/vnd.ms-htmlhelp`, `application/mshelp` and `chemical/x-chemdraw`. I strongly assume the last one is the one I was actually supposed to use.

Thanks for pointing me to it!

PR #830, opened 2025-11-30
Use default credentials in RDP test

cpiber:
It's needlessly confusing and annoying to use different credentials for this one command compared to everywhere else


oskardotglobal (approved):
What a weird way for the docs to say this. Thanks for the catch

PR #832, opened 2025-12-01
Add Dorico 6

basil:
I tested this by discovering and launching my copy of Dorico 6.


LDprg:
Just for reference #834, regarding the generic icons.

PR #837, opened 2025-12-08
fix: arch linux has deprecated gnu-netcat

ZhymabekRoman:
See https://gitlab.archlinux.org/archlinux/packaging/packages/gnu-netcat/-/issues/1
gnu-netcat only available in AUR: https://aur.archlinux.org/packages/gnu-netcat


oskardotglobal:
Ah, this seems to have not been changed with #528, we have found openbsd-netcat to work best

ZhymabekRoman:
Oh, should I then change netcat to openbsd-netcat?

oskardotglobal:
Yes, please

ZhymabekRoman:
done

Issue #827, opened 2025-11-23
Random PowerShell windows when opening apps


TheChilliPL:
### What happened?

Often when opening apps via winapps, a random PowerShell window also appears along with the app. The PowerShell window is often empty or almost empty. Seems to only happen the first time opening the app since the last reboot

### Your FreeRDP version and where you got it from

FreeRDP version: 2.11.7

### Your Linux distribution and version

EndeavourOS x86_64 + 6.17.6-arch1-1

### Your `winapps.conf`

```shell
RDP_USER="Chilli"
RDP_PASS="MY_PASSWORD_HERE"
RDP_IP="127.0.0.1"
WAFLAVOR="docker"
RDP_FLAGS="/cert:tofu /sound /microphone +home-drive"
AUTOPAUSE="on"
```

### Logs

```shell

```

### Terms

- [x] I am running the latest version.
- [x] To the best of my knowledge, this is a bug and not a setup nor a FreeRDP problem.
- [x] I have checked for duplicate issues.
- [x] I agree to follow this project's Code of Conduct.

CaptainException:
I had the same issue. Sometimes I even had multiple windows spawn. The culprit was a scheduled task that tried to run the script `C:\Windows\TimeSync.ps1` which did not exist. Once you disable that scheduled task it won't happen again.

P.S.: you can use the [Autoruns](https://learn.microsoft.com/en-us/sysinternals/downloads/autoruns) utility to monitor all startup tasks on windows (task manager only shows a very small subset)

oskardotglobal:
That script should exist, it's used to ensure the container time doesn't go out of sync.

PR #870, opened 2026-01-24
docs(libvirt-install):Add TimeSync.ps1 Download

doudou0720:
Should Solve #827 

The reason is so simple:no one ask them to download TimeSync.ps1


doudou0720:
Sorry for not sign-off cause I edit it on Github Mobile)

PR #874, opened 2026-01-25
Improve FreeRDP process lifecycle tracking and cleanup

### LINKED ISSUE #252: Winapps does not output any usage help when run without any arguments
@oskardotglobal found during testing of #234, see #234 for more details.

KernelGhost:
This PR improves FreeRDP process lifecycle tracking, with better support for the Flatpak as well as automatic cleanup of orphaned process tracking files on startup. This PR also has the potential to address https://github.com/winapps-org/winapps-launcher/issues/27.

> [!WARNING]  
> *I no longer have WinApps installed, so I’ve only been able to validate this through basic unit tests. I’d appreciate another maintainer testing the changes. Many thanks in advance.*

Closes #252.

### DIFF
--- a/bin/winapps
+++ b/bin/winapps
@@ -63,13 +63,13 @@ FREERDP_PID=-1
 NEEDED_BOOT=false
 
 ### TRAPS ###
-# Catch SIGINT (CTRL+C) and SIGTERM to call 'waCleanUpAndExit'.
-trap waCleanUpAndExit SIGINT SIGTERM
+# Catch SIGINT (CTRL+C) and SIGTERM to call 'waKillCleanExit'.
+trap waKillCleanExit SIGINT SIGTERM
 
 ### FUNCTIONS ###
-# Name: 'waCleanUp'
+# Name: 'waCleanFreeRDP'
 # Role: Clean up remains prior to exit.
-function waCleanUp() {
+function waCleanFreeRDP() {
     for proc_file in "${APPDATA_PATH}"/FreeRDP_Process_*.cproc; do
         # Protect against glob non-expansion.
         [[ -f "$proc_file" ]] || break
@@ -106,7 +106,6 @@ function waCleanUp() {
 function waKillFreeRDP() {
     # Declare variables.
     local TERMINATED_PROCESS_IDS=()
-    local TERMINATED_PROCESS_IDS_STRING=""
 
     # Loop through each matching file and add to the array.
     for FREERDP_PROCESS_FILE in "${APPDATA_PATH}/FreeRDP_Process_"*.cproc; do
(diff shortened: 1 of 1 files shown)


Maxb0tbeep:
I just tested this on Arch Linux (not flatpak) and it works as intended. I made a duplicate `.cproc` file so there would be an orphan file at `~/.local/share/winapps` and upon connecting to a FreeRDP session it removes the orphaned file. However, it might be better (if possible) to remove orphaned files on WinApps startup for the edge case of a user experiencing the orphaned `.cproc` file issue and then being unable to gracefully power off without opening a FreeRDP session. I can see this happening like this:
1. Host computer power loss (`.cproc` doesn't get removed) 
2. The user turns their computer back on and WinApps autostarts or is manually started 
3. The user never opens a FreeRDP session but the container/VM is running
4. The user eventually wants to power off, hibernate, pause, etc. and the operation fails

Futhermore, an orphaned `.cproc` file also prevents auto pausing which I'm sure almost anyone who would autostart WinApps would have enabled.

KernelGhost:
@Maxb0tbeep Thanks for taking the time to test these changes! In the commits above, `waCleanUp` is run on script startup, so orphaned files should be cleaned up automatically when a new instance of winapps is launched. Is this sufficient to address your concerns?

Maxb0tbeep:
@KernelGhost Interesting. The behavior that I experienced when I cloned your branch was that it only cleaned up orphaned files once a FreeRDP session was started, so starting up WinApps from the launcher (or just starting the container manually) didn't remove the files.

KernelGhost:
@Maxb0tbeep That behaviour is expected given the way things are currently written. Implementing cleanup at container launch would be difficult, but your described desired behaviour can certainly be added to the launcher.

To avoid having two separate projects (winapps and winapps-launcher) modifying process-lifecycle tracking files, one possible approach would be to add a new winapps command (something like 'killrdp') to the main winapps script. This command could inherit the existing FreeRDP termination logic currently present within the launcher, while also explicitly handling cleanup of the cproc files (currently not done by the launcher implementation).

The launcher would then invoke an instance of winapps with the 'killrdp' argument rather than implementing its own termination logic.

Curious to hear your thoughts on whether this approach would be acceptable?

Maxb0tbeep:
@KernelGhost sounds good to me.

KernelGhost:
@Maxb0tbeep Can you please test these changes?

Maxb0tbeep:
@KernelGhost works for me :)

KernelGhost:
@Maxb0tbeep Thanks for taking a look.

@oskardotglobal would you mind proofreading and testing this alongside https://github.com/winapps-org/winapps-launcher/pull/43 as a final check?

KernelGhost:
@oskardotglobal Any updates on this?

oskardotglobal:
@KernelGhost my bad, I‘ve been busy, I‘ll take a look later today

oskardotglobal (approved):
I've tested this and can confirm it works (at least when I manually kill winapps or reboot)

PR #882, opened 2026-02-17
Include freerdp in Nix package instead of weird alias

oskardotglobal:
This can cause weird issues causing WinApps to think FreeRDP is not installed.

### DIFF
--- a/packages/winapps/default.nix
+++ b/packages/winapps/default.nix
@@ -7,7 +7,6 @@
   libnotify,
   netcat,
   iproute2,
-  writeShellScriptBin,
   nix-filter ? throw "Pass github:numtide/nix-filter as an argument!",
   ...
 }:
@@ -30,7 +29,7 @@ stdenv.mkDerivation rec {
 
   nativeBuildInputs = [ makeWrapper ];
   buildInputs = [
-    (writeShellScriptBin "xfreerdp3" ''${lib.getExe' freerdp "xfreerdp"} "$@"'')
+    freerdp
     libnotify
     dialog
     netcat


Letgamer:
@AkechiShiro do you have knowledge about this?
There was probably a reason it was done this way

oskardotglobal:
I did this. I don't know why, though.

AkechiShiro:
I'd like to add a test but it's complex, I was thinking of using NixOS test framework spawn a VM pull Windows 11 smallest image and RDP inside after installing Winapps but I'm not sure it's easy to do that as I don't see how to use the test framework outside of NixOS module and outside of Nixpkgs.

Would this be worth it on the legacy version too, maybe we should focus on testing the Rust rewrite properly ? Any input on this idea ? @oskardotglobal @LDprg

oskardotglobal:
I was thinking about this too, but I don‘t know if that‘s even feasible in CI, since there would have to be a window created on the runner. It would be really cool if it worked though.

oskardotglobal:
Also regarding the rewrite, feel free to review my PR and give your thoughts

LDprg:
I agree that would be very nice (I don't know if it is worth to implement for legacy, however I didn't contribute to the project as much due to very much to do in private life sadly).

I am not sure whether github action bandwidths (networking and cpu time) allow us to do so, however we could implement a cache so a windows image gets tested directly, instead of downloaded every time (maybe there is even a feature for this). Still I am not sure about actions interface limitations (like can docker or podman even run)?

oskardotglobal:
we can definitely run the VM inside actions, the biggest issue is probably that the test would take quite some time and how to get usable output from the rdp session working, something like screenshots?

AkechiShiro:
I think we would need to investigate if it's possible to start a fake X11 server or Wayland session where we spawn the RDP client window and we attach a screenshot tool to it to check that we do indeed have a working login in a Windows machine or a Windows machine installation starting.

Podman or Docker is possible to use in Actions @LDprg, here is Quickemu's vm test for instance : https://github.com/quickemu-project/quickemu/blob/master/.github/workflows/test-build-quickemu.yml#L47-L81

Maybe instead of using RDP, we expose an SSH port from the Windows in order to verify that it is running and working properly, we use ssh, although launching `winapps` is better for a test of `winapps`.

oskardotglobal:
SSH is already planned for the rewrite, I've almost completely replaced all uses of freerdp to spawn cmd.exe with it.

oskardotglobal:
We should really discuss testing in a new thread though

PR #900, opened 2026-03-16
Do not expose RDP to the local network by default

oskardotglobal:
As discussed in #897


gamer191:
Thanks

Now that I think about it, does the exposed web interface require a security advisory, especially since it requires users to make a manual change?

On the other hand, it’s technically an issue with the documentation and not with the actual software

And should I report this upstream to dockur/windows? From a quick look it looks like their readme example also exposes the web interface

Nullum-Idea (approved):
LGTM

PR #936, opened 2026-06-02
[ci] Drop garnix

oskardotglobal:
Our nix CI provider garnix is ending their service. This PR replaces it with a github actions workflow.


LDprg (approved):
Seem fine.

PR #938, opened 2026-06-07
Proposal to add some text to README on how to manage apps after installation using the WinApps Install Wizard as end user documentation

alexrhyshurn:
I propose to add a section to the README giving some end user information about how to manage apps after installation using the WinApps Install Wizard, and detail the options that are available to ./setup.sh

I feel that a section like this will be clearer for end users and reduce user support issues being raised.

Please modify/add/reject as you see fit.

Thank you for your consideration.
Alex

ps: Forgive any mistakes, as this is my first pull request ever in my life :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -672,7 +672,7 @@ Usage:
       --help                                        # Display this usage message.
 ```
 
-To rerun the app detection go to the installation location and run the WinApps Install Wizard:
+To rerun the app detection go to the installation location and run the WinApps Install Wizard as the user that WinApps is installed for:
 ```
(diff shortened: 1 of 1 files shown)


oskardotglobal (approved):
Thanks. I've done some rewording but this is a useful addition. By the way, you should be able to run the setup using `winapps-setup`, without changing directories.

alexrhyshurn:
Thank you, and now I learned about the winapps-setup command instead of ./setup.sh

Thanks again for the great project.

PR #951, opened 2026-07-27
fix(autopause): stop container on idle instead of pausing

QwertyJacob:
### Summary of Changes

This PR updates the `AUTOPAUSE` functionality in `bin/winapps` to stop (`docker compose stop` / `podman-compose stop` / `virsh shutdown`) idle virtual machines rather than pausing (`docker compose pause` / `virsh suspend`).

### Problem Addressed

When `AUTOPAUSE="on"` triggers after the idle timeout:
1. Previously, `winapps` executed `docker compose pause`. In Docker, `pause` uses Linux cgroups freezer to freeze process execution in memory. This keeps the container listed as `Up (Paused)` in `docker ps` and leaves all host RAM allocated.
2. More critically, for QEMU/KVM virtual machines (such as `dockur/windows`), freezing QEMU with cgroups freezer breaks clock/timer synchronization between host KVM and guest vCPUs. This causes QEMU helper threads to spin continuously at ~150% host CPU load in kernel space, causing severe host CPU overheating and maximum fan speeds.

### Solution

- Replaced `pause` with `stop` (`docker compose stop`, `podman-compose stop`) and `virsh shutdown` when `SUSPEND_WINDOWS` is triggered in `waCheckIdle`.
- Updated notification text to: `Stopping Windows due to inactivity.`
- When `stop` is called, the VM gracefully shuts down, dropping host CPU usage to 0% and releasing host RAM completely.
- `winapps` already natively handles booting from an `exited` container status in `waCheckContainerRunning` (`docker compose start`), so next time a user opens any WinApp shortcut, Windows boots up automatically and transparently.


oskardotglobal:
I'll have to investigate this. Can you please

- sign off your commit
- confirm that you experienced the overheating issue personally
- and confirm whether this pull request was AI-generated in accordance to our Contribution Policy

QwertyJacob:
Hi Oskar,

Thanks for answering and for your patience with this late answer.
I did not measured temperature whatsoever, I just heared my fan gone crazy, then I confirmed my CPU was doing a lot of work. Does that make sense? I guess the main problem is that, under my conditions "pausing" the container provoke that. Is that something you do not experience? 

I do not pretend you answers, you've already done too much for me :) thanks. Just wanted to help... Ah and the PR was AI-generated, I've controlled the content though...

Cheers,
Jesus

oskardotglobal:
According to https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/virtualization_administration_guide/sub-sect-starting_suspending_resuming_saving_and_restoring_a_guest_virtual_machine-suspending_a_guest_virtual_machine, `virsh` will behave properly. So on virsh, we should always pause and maybe add the option to shut down (if the user needs the RAM). 

The docker/podman changes are fine for now, though there should be a better long-term fix; I will ask the folks at dockur/windows about this.

QwertyJacob:
Absolutely,
Thanks for answering,
Cheers,
J

PR #956, opened 2026-08-09
Fix connection for podman with pasta networking

blegat:
On ArchLinux with podman version 6.0.1. The Word app wasn't launching. It seems to be because of the new pasta networking with modern podman version.
To be clearn, consider the three places:
1. The linux desktop (**host**).
2. The rootless network namespace : the **sandbox** `podman unshare --rootless-netns` steps into.
3. The **container** itself.

For an old podman (slirp4netns era), the port forwarder was reachable at `127.0.0.1:3389` inside the **sandbox**.
For a modern podman (pasta/netavark era), the forwarder (rootlessport) binds `127.0.0.1:3389` in the **host**.

So I first tried using `127.0.0.1:3389` directly from the **host**. The **sandbox** then redirects to the **container**'s 127.0.0.1. However, it seems that  and Windows VM's port forwarder inside the container (passt/dnsmasq from dockur/windows) only accepts traffic arriving through the container's network card (eth0), not its loopback. Worse, it always accepts and only fails afterward (so a simple "is the port open?" test says yes even when it's broken).

What works for me is to step inside the sandbox and dial the container's real IP address instead of 127.0.0.1. That traffic enters the container through eth0 and it works.

With this PR, we first try connecting through the container ip. If it fails, we fallback to what was done before this PR.


oskardotglobal:
Hi, thanks for the PR. Please fix the errors reported by CI and sign off on your commits.

blegat:
All green now :)

oskardotglobal (approved):
Thanks.

Issue #825, opened 2025-11-19
HiDPI / fractional scaling support per-monitor


oskardotglobal:
### Discussed in https://github.com/winapps-org/winapps/discussions/824

<div type='discussions-op-text'>

<sup>Originally posted by **mlltAero** November 19, 2025</sup>
Currently, WinApps uses a global RDP_SCALE for all RemoteApp windows. On HiDPI laptops, this causes RemoteApps to render tiny because the scale does not automatically match the display’s effective scaling.

### Problem:

RDP_SCALE is static and does not account for fractional scaling in GNOME or per-monitor DPI.
Users with HiDPI laptop screens must manually adjust RDP_SCALE in winapps.conf.

<img width="2559" height="1599" alt="image" src="https://github.com/user-attachments/assets/2828ccaa-9dbf-40b9-a353-5360c2c610b9" />

**Proposed feature:**

Automatically detect the primary monitor and its scaling (e.g., GNOME fractional scaling or via xrandr resolution vs. native panel size) and convert to supported scaling factor (nearest neighbor).
Adjust RDP_SCALE for RemoteApp windows on launch so they render correctly on HiDPI displays.
</div>

DonTMover:
in gnome 49 on hidpi laptop +

abirkel:
I think winapps needs to expose this xfreerpd arg `/scale-desktop: <percentage>      Scaling factor for desktop applications`.

It is currently exposing only this arg `/scale: [100|140|180]             Scaling factor of the display` in the config file.

butterdori:
Same issue for me. If even one of my connected monitors is not at 100% scale, the window becomes much smaller.
I can solve this by either 1) setting scale 100% to all monitors or 2) set rdp_scale to 180 in winapps.conf
For reference, Gnome 49.3 on ThinkPad T14 laptop.
Laptop monitor: 1920x1200, scale 100%
External monitor 1: 1920x1080, scale 100%
External monitor 2: 2560x1600, scale 125% <-- this is the culprit for me.

PR #960, opened 2026-08-21
feat: pass the exact display scale to FreeRDP, and detect it automatically

oraturk75:
## What this changes

`RDP_SCALE` was rounded to 100, 140, or 180 because WinApps assumed FreeRDP only supported those values. FreeRDP takes scaling values three different ways.

`/scale-desktop`: Can take any integer between 100-500, and this is where full-desktop and RemoteApps scaling is set.

`/scale-device`: Accepts only 100, 140, and 180. This is the scaling factor for MS Store applications.

`/scale`: This is only a convenience shortcut. It takes one value and assigns it to the above two. It accepts only 100, 140, and 180. This is what WinApps is currently using. Hence, a user requested scale of 125 is rounded to 140 by WinApps before being sent through FreeRDP’s `/scale` option.

This PR now sends the exact scaling value through `/scale-desktop`, calculates/rounds the corresponding value, and then sends it to `/scale-device` separately. I had to set both, not only `/scale-desktop`, to keep the existing behaviour unchanged.

As requested, `RDP_SCALE` now controls RemoteApp launches, while a new `RDP_DESKTOP_SCALE` controls the full-desktop session. When the user does not set `RDP_DESKTOP_SCALE`, it takes the value of `RDP_SCALE` by default.

Both settings also accept `auto`, which is now the default. This implements the automatic host scale detection part of #825 for a single-monitor setup.

## Configuration

```sh
RDP_SCALE="auto"
RDP_DESKTOP_SCALE="${RDP_SCALE}"
```

The name `RDP_DESKTOP_SCALE` refers to the full-desktop launch type, not to be confused with FreeRDP’s `/scale-desktop` option. Both WinApps settings provide a `/scale-desktop` value for their own respective launch type and automatically calculate an acceptable (100, 140, or 180) `/scale-device` value.

Existing configurations do not need to change. If `RDP_SCALE` is missing, it uses `auto`. If the new `RDP_DESKTOP_SCALE` is missing, it takes the resolved `RDP_SCALE` value.

## Automatic detection

On Cinnamon/X11, `auto` reads the primary monitor’s applied scale from Muffin’s `DisplayConfig`.

Automatic detection is designed to do its best, but fail safely and never block a launch. If detection is unavailable or fails, WinApps falls back to 100 and records the reason in its own log. The user can always bypass detection by manually setting their own desired value.

I did not implement per-monitor scaling for multi-monitor setups, rescaling an actively running RemoteApp when it moves between monitors, monitor hotplug, or scale changes during an active session.

The detection is kept in a separate function, so support for other desktop environments can be added without changing the configuration or launch code.

WinApps now validates the values before sending them; non-integers fail with a visible error, values below 100 are changed to 100, and values above 500 are changed to 500.

## Testing

Tested against Win11 Pro, FreeRDP 3.30.0, and Cinnamon 6.4.13 under X11. No test framework was added.

**Exact scale:** I tested `winver` as a RemoteApp using the geometry reported by my Win11 VM. Scale 100 produced 460x423, 125 produced 575x521, and 150 produced 691x621. The dimensions followed the requested scale within two pixels.

**Automatic detection:** With my host at 100%, WinApps detected 100. After setting my host to 175% in Cinnamon, WinApps detected the applied value of 176. After restoring my host to 100%, WinApps detected 100 again. I also tested automatic detection from Cinnamon’s application menu, and it returned 176.

**Detection failures:** I tested every detection failure path, including missing `gdbus`, a failed D-Bus call, malformed replies, missing primary-monitor information, and invalid detected scales. Every case fell back to 100 without blocking the launch.

**End-to-end scaling:** With my host at 175%, a Windows test script inside the RemoteApp session reported 168 DPI. This confirms that the detected scale reached the Windows session.

**Launch mapping:** I captured the FreeRDP commands generated by all four WinApps launch paths. The full-desktop launch used the full-desktop scale, while all three RemoteApp launches used the configured `RDP_SCALE` value. When no separate full-desktop value was configured, the full-desktop launch also used `RDP_SCALE` value.

`bash -n`, ShellCheck, `git diff --check`, and the full `pre-commit` suite all pass.

**Not tested with Wayland.** WinApps uses `sdl-freerdp3` for full-desktop launches on Wayland, while my environment uses X11.

## AI Assistance Disclosure

I got help from two AI assistants to write most of the code. I reviewed their work, made the final decisions, and personally tested and confirmed the intended functionality.


oskardotglobal:
So if I understand this correctly, if I set a scale of 158, this will scale the desktop accordingly, but RAILS will still be scaled to 140? That seems unintuitive. Maybe we should just have a seperate scale for desktop and RAILS?

oraturk75:
Good question, and the README wording is what made it ambiguous. Sorry about that.

`/scale-device` is not the RemoteApp scale. Per MS-RDPBCGR 2.2.1.3.2 these are two distinct fields with different allowed ranges, sent together and each validated against the other: `desktopScaleFactor` is valid from 100 to 500, and `deviceScaleFactor` is restricted to 100, 140 or 180. `/scale-desktop` is what sizes the session, and that includes RemoteApp windows.

I measured this before writing the patch. Winver under RemoteApp, with `/scale-device` pinned at 100 and only `/scale-desktop` varied, using the window geometry the server reports:

| `/scale-desktop` | window | width ratio |
|---|---|---|
| 100 | 460 x 423 | 1.000 |
| 125 | 575 x 521 | 1.250 |
| 150 | 691 x 621 | 1.502 |

Linear fits on both axes are within 2 px across the three points, with a fixed ~31 px offset on height that is window chrome. So at `RDP_SCALE=158` the RemoteApp windows scale at 158, not 140.

The old `/scale` set both fields. Sending only `/scale-desktop` would silently change `deviceScaleFactor` from the rounded value WinApps currently sends to FreeRDP's default of 100, so this patch preserves the existing device-factor behavior. That field is the only value being approximated.

I will clarify the README. Does this address your concern, or would you still prefer separate user control of `deviceScaleFactor`?

oskardotglobal:
I think a seperate config option for the desktop scale defaulting to the value of RDP_SCALE would be preferred, with RDP_SCALE controlling RemoteApp scaling.

Also, please do not copy-paste the answer of the LLM as the response to my question. As per our policy, we accept AI-generated code, but this is just bad etiquette and doesn't really support your claim of having reviewed the changes by hand.

oraturk75:
Hey, I thought more about this PR work and #825, which inspired part of it, and I believe something got lost in the transit. There is no reason why I should not include automatic detection of the primary monitor scale (on Cinnamon/X11) from #825 as part of this PR, since it completes the scaling work already here. I do not have a multi-monitor setup, so I cannot reproduce nor properly test those issues and will leave that part of #825 for someone else.

I will add the separate full-desktop scale option per your request. Also; I have decided to expand this PR to include automatic scale detection. Of course, manual scale settings will always override automatic detection. Is that okay with you?

PS: Sorry about earlier posts, didn't mean any disrespect.

oskardotglobal:
Sure, automatic scale detection sounds super useful! Also, multi-monitor scaling is a non-trivial issue and FreeRDPs multi-monitor capabilities are dubious anyways, so focusing on single-monitor scale detection seems like the right choice

oraturk75:
@oskardotglobal Pushed. `RDP_SCALE` now controls RemoteApps and the new `RDP_DESKTOP_SCALE` controls the full desktop, defaulting to `RDP_SCALE` as you asked. Both accept `auto`, which detects the primary monitor scale on Cinnamon/X11. If detection is unavailable or fails, it falls back to 100, while manual values always override detection. I also clarified the README as promised and updated the title and description for the new scope.

oraturk75:
@oskardotglobal Hey, just checking in. Did you had a chance to look at the latest changes?
Please let me know if anything else is needed from me.

oskardotglobal:
Hi, I still need to find some time to review this properly. Will get back to you

oskardotglobal (changes requested):
After giving this a proper review: this is not something I will merge as-is anytime soon. The long inline awk code is iff and for fact it only supports Cinnamon on X11 it's implemented way and documented way too broadly; until this supports more setups this won't be very useful to many people. Also, the AI-generated prose in the documentation & comments is poorly phrased.

oraturk75:
@oskardotglobal It is hard for me to hit a moving target. I said Cinnamon/X11 up front and you said "sure". So that part isn't new. Also, “poorly phrased” is not actionable, it's “way too broadly”.

Back to work.

I will do better with README and Code Documentation next time. I realize that it is too long, to say the least. No need to point where and what's wrong with the current docs now; they're getting rewritten anyway.

Only `auto` is DE-specific. The rest of the PR is the same on any desktop.

I can also support GNOME/Wayland through Mutter, which exposes the same monitor data as Cinnamon's Muffin. I can test GNOME/Wayland locally myself, but a second test from someone who uses GNOME as their default DE would be better. Is that enough, or which other environments are required?

I have a pretty good idea how to make the `awk` code much better, likely without the inline parser at all, but it will wait until your new requirements are clear.

Please let me know. Thanks.

oskardotglobal:
Sorry, I missed the "Cinnamon/X11" part of your original comment...

What I meant by "implemented too broadly" is that for example, for something implemented currently only for Cinnamon it shouldn't be the default behaviour; the default behaviour should be to configure the scale numerically and then the user can opt-in to autoscaling if they're on a supported DE.

`waFixScale` is getting a bit complex and the newly added 21 (!) local variables aren't helping. Also, since `waDetectHostScale` is pretty DE-specific, it should probably detect the DE first and then call a `waDetect<DE>HostScale`.

oraturk75:
Understood on all four.

I will make the default numeric 100 and keep `auto` opt-in.

I will simplify `waFixScale`, split `waDetectHostScale` into DE detection plus `waDetect<DE>HostScale`, and replace the long inline `awk` block with a simpler shared parser called by the DE-specific detectors.

I tested GNOME/Wayland, and Mutter-based detection is feasible. Do you want it in this PR or in a follow-up?

I will rewrite the docs to match the final behaviour.

oskardotglobal:
Put GNOME in a follow-up, take your time.

PR #961, opened 2026-08-23
Add Support for Magic the Gathering Online

jam1garner:
First off: great project! Have been wanting (and vaguely considering making) something along these lines for a while, really glad to have learned about it.

This adds support for the Magic the Gathering Online Client, which is a piece of software which runs terribly under other solutions and is kinda annoying to spin up a dedicated VM for. Interestingly it's a ClickOnce application and basically everything I tried to implement that without adding a built-in configuration for didn't work (the most malware-looking option was base64-encoded powershell being run through `winapp manual powershell.exe -EncodedCommand ...`), so upstreaming it so that I can recommend winapps for this purpose to other MTGO+Linux users.

Icon is a vectorized trace of the official icon, but if you'd prefer the actual official icon (identical minus shading) I can extract and include that (although it will probably just be a bitmap image embedded in the SVG, if that matters).

Let me know if there are any other changes needed!


apps/mtgo/info
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2024 Fmstrat
+# All rights reserved.
+#
+# SPDX-License-Identifier: Proprietary
+
```
oskardotglobal: Please remove this header as per CONTRIBUTING.md

apps/mtgo/info
```diff
@@ -0,0 +1,19 @@
...
+# GNOME shortcut name
+NAME="MTGO"
+
+# Used for descriptions and window class
+FULL_NAME="Magic the Gathering Online"
+
+# The executable inside windows
+WIN_EXECUTABLE="C:\Users\%USERNAME%\AppData\Roaming\Microsoft\Windows\Start Menu\Programs\Daybreak Game Company LLC\Magic The Gathering Online .appref-ms"
```
oskardotglobal, line 8: Is the space before the `.appref-ms` intended?

oskardotglobal (changes requested):

apps/mtgo/info
jam1garner: Yes, it bugs me too but it's in the real thing

jam1garner:
Apologies for missing the CONTRIBUTING.md, I had looked for one but wasn't aware of the `.github/readme` structure being a thing, sorry. Will also add that I agree to the DCO, all changes are my own and I'm happy to license them according to the AGPLv3.

oskardotglobal (approved):

PR #968, opened 2026-09-08
Resolve relative paths to absolute before Windows path conversion

miles-read:
Small change to fix opening files using relative paths.

The existing sed substitution only matches paths already prefixed with $HOME or $REMOVABLE_MEDIA, so a relative path falls through unconverted. 

```bash
[ "${2#/}" = "${2}" ] && set -- "${1}" "$(pwd)/${2}"
````


oskardotglobal:
Hi, please sign off on your commits (see the DCO check on how to do that). Also, please indent the if statement and the comment by another 4 spaces. Then, this will be ready to merge.

oskardotglobal:
Your first commit still isn't signed off.

